PR #8, opened 2018-08-07
combined upstream merge from BI X production

clemensutschig:
upstream commit from getting ods work on premise


clemensutschig:
@michaelsauter - I guess there is something more to do ... 

[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
ERROR: This job: prov-cd/prov-cd-prov-app-test is not responsible for building: production

the pipeline points now to @production - but building test against test fails with the above

PR #56, opened 2018-11-12
Feature/springboot upgrade to 1.5.15

### LINKED ISSUE #55: upgrade provision app to spring boot v1 latest (1.5.15)
fix for ODS v1 - fup from  #39

clemensutschig:
fixes #55

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,8 @@
 buildscript {
 	 
     ext {
-        springBootVersion = '1.5.9.RELEASE'
+        springBootVersion = '1.5.15.RELEASE'
+		// springBootVersion = '2.0.5.RELEASE'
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -1,7 +1,8 @@
 buildscript {
 	 
     ext {
-        springBootVersion = '1.5.9.RELEASE'
+        springBootVersion = '1.5.15.RELEASE'
+		// springBootVersion = '2.0.5.RELEASE'
```
michaelsauter: Could we remove the comments about version 2.0.5 in this file?

Would be nice to have this clean - plus by the time we upgrade, we might already go to another version (e.g. we use 2.1.0 already).

PR #61, opened 2018-11-21
Create new webhooks

michaelsauter:


michaelsauter:
@clemensutschig Yes, I tested end-to-end in the RedHat-managed OpenShift cluster. I tried to provision a new project and it worked. 

So the only thing I have not tried is to test this end-to-end on a completely new OpenShift cluster ...

PR #62, opened 2018-11-29
Unify, simplify and update OCP config

michaelsauter:
Instead of having separate config for prov-dev and prov-test, we now
have one config, with different parameters. Also, the prov-cd config has
been adjusted to include the new webhook proxy.


clemensutschig:
@michaelsauter  - please ensure that you change the getting started gzuide as well.

PR #72, opened 2018-12-13
Provision app - add admin if passed to rdeck call and archive of jenkinsfile

### LINKED ISSUE #71: Openshift projects are created with current user as admin, even in case of special permission set
Today we pass down the current (logged in) principal to rundeck and subsequently to openshift (as project role admin) example below

tech_integration          admin   -> the currently logged in user 

rather than taking the provided project admin (in case of special permission set = true)

https://github.com/opendevstack/ods-provisioning-app/blob/master/src/main/java/org/opendevstack/provision/services/RundeckAdapter.java#L185

clemensutschig:
fixes #71


Jenkinsfile
```diff
@@ -28,6 +28,7 @@ odsPipeline(
   stageScanForSonarqube(context)
   stageStartOpenshiftBuild(context)
   stageDeployToOpenshift(context)
+  stageCollectBuildArtifacts(context)
```
michaelsauter: Let's remove this from the PR, because once https://github.com/opendevstack/ods-jenkins-shared-library/pull/52 is merged, the latest master of the prov app will automatically have it (and then would do it twice ...)

clemensutschig:
1.x fix - closing and opening a new one

PR #74, opened 2018-12-17
Bug/oc project admin & description length - fix for 1.0x

### LINKED ISSUE #71: Openshift projects are created with current user as admin, even in case of special permission set

clemensutschig:
fixes #71 against 1.0x
fixes #70 against 1.0x

### DIFF
--- a/src/main/java/org/opendevstack/provision/services/RundeckAdapter.java
+++ b/src/main/java/org/opendevstack/provision/services/RundeckAdapter.java
@@ -32,6 +32,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.core.userdetails.UserDetails;
 import org.springframework.stereotype.Service;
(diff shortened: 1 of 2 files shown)


clemensutschig:
@michaelsauter  - i sort of understand you.. but its important enough while surgical to do imho..

michaelsauter:
@clemensutschig I don't want to be a pain, but this change is somewhat breaking - you need to rebuild the base slave image, and the maven slave image for it to work. I would want to avoid this for patch releases unless it is 100% required to fix a bug. Here, we are dealing with a feature. 

Since you already have this on prem in the `production` branch, I don't see a benefit to put this into `1.0.x` either ...

clemensutschig:
@michaelsauter- you are.. but thats ok. Decline the sqjar One. And that one i fix. Yak

michaelsauter:
@clemensutschig I rewrote the branch - do the changes look good to you now?

PR #79, opened 2019-01-24
Feature/pick project template & add attachment permission

### LINKED ISSUE #26: Provision app should support different (jira/confluence) project types
today - the template used for jira & confluence is hardcoded (thru a config file/map) and for jira in the source

`#Data for confluence space creation
confluence.blueprint.key=com.atlassian.confluence.plugins.confluence-software-project:sp-space-blueprint`

 This should be at least configurable

### LINKED ISSUE #78: A jira project does not have the ability to add an attachment to an issue
I've just noticed than the default permissions on a jira project after the provisioning app, does not have the ability to add an attachment to an issue, could be the possibility to do this by default, -manager and -team

clemensutschig:
fixes #26 and fixes #78

### DIFF
--- a/src/main/java/org/opendevstack/provision/controller/DefaultController.java
+++ b/src/main/java/org/opendevstack/provision/controller/DefaultController.java
@@ -15,6 +15,7 @@
 package org.opendevstack.provision.controller;
 
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import javax.servlet.http.HttpServletRequest;
@@ -71,6 +72,9 @@ public class DefaultController {
   @Value("${openshift.project.upgrade}")
   private boolean ocUpgradeAllowed;
   
+  @Autowired
+  List<String> projectTemplateKeyNames;
+  
   @RequestMapping("/")
   String rootRedirect() {
     return "redirect:home.html";
@@ -95,6 +99,7 @@ public class DefaultController {
             model.addAttribute("crowdUserGroup", crowdUserGroup);
             model.addAttribute("crowdAdminGroup", crowdAdminGroup);
             model.addAttribute("ocUpgradeAllowed", ocUpgradeAllowed);
+            model.addAttribute("projectTypes", projectTemplateKeyNames);
         }
         model.addAttribute("classActiveNew", ACTIVE);
         return "provision";
--- a/src/main/java/org/opendevstack/provision/controller/ProjectApiController.java
+++ b/src/main/java/org/opendevstack/provision/controller/ProjectApiController.java
@@ -68,7 +68,7 @@ public class ProjectApiController {
   private static final String STR_LOGFILE_KEY = "loggerFileName";
   
   @Autowired
-  private JiraAdapter jiraAdapter;
+  JiraAdapter jiraAdapter;
   @Autowired
   private ConfluenceAdapter confluenceAdapter;
   @Autowired
@@ -88,6 +88,9 @@ public class ProjectApiController {
   @Autowired
   private IProjectIdentityMgmtAdapter projectIdentityMgmtAdapter;
 
+  @Autowired
+  private List<String> projectTemplateKeyNames;
+  
   // open for testing
   @Autowired
   CustomAuthenticationManager manager;
@@ -374,6 +377,13 @@ public class ProjectApiController {
     return ResponseEntity.ok().build();
   }
 
+  @RequestMapping(method = RequestMethod.GET, value = "/templates")
+  public ResponseEntity<Object> getProjectTemplateKeys(HttpServletRequest request,
+      @CookieValue(value = "crowd.token_key", required = false) String crowdCookie) 
+  {
+    return ResponseEntity.ok(projectTemplateKeyNames);
+  }
+  
   /**
    * Validate the project's key name. Duplicates are not allowed in JIRA.
    * 
--- a/src/main/java/org/opendevstack/provision/services/ConfluenceAdapter.java
+++ b/src/main/java/org/opendevstack/provision/services/ConfluenceAdapter.java
@@ -30,6 +30,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
+import org.springframework.core.env.Environment;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
 import org.springframework.stereotype.Service;
@@ -90,6 +91,15 @@ public class ConfluenceAdapter {
   @Value("${global.keyuser.role.name}")
   private String globalKeyuserRoleName;
   
+  @Autowired
+  Environment environment;
+  
+  @Autowired
+  List<String> projectTemplateKeyNames;
+  
+  @Value("${project.template.default.key}")
+  private String defaultProjectKey;
+  
   public ProjectData createConfluenceSpaceForProject(ProjectData project, String crowdCookieValue)
       throws IOException {
     this.crowdCookieValue = crowdCookieValue;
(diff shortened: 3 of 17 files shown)


src/main/java/org/opendevstack/provision/services/ConfluenceAdapter.java
```diff
@@ -161,17 +171,36 @@ protected String getJiraServerId() throws IOException {
...
         getList(url, crowdCookieValue, new TypeReference<List<Blueprint>>() {});
+    
+    String confluencetemplateKeyPrefix = "confluence.blueprint.key.";
+
+    String template =
+    	(projectTypeKey != null && !projectTypeKey.equals(defaultProjectKey) && 
+    		environment.containsProperty(confluencetemplateKeyPrefix + projectTypeKey) && 
+    		projectTemplateKeyNames.contains(projectTypeKey)) ?
+    		environment.getProperty(confluencetemplateKeyPrefix + projectTypeKey) : 
+    			confluenceBlueprintKey;
```
michaelsauter: Maybe we can extract the condition, or even returning the whole template string into a private method, I find this hard to read ...

src/main/java/org/opendevstack/provision/services/JiraAdapter.java
```diff
@@ -108,9 +123,28 @@ public ProjectData createJiraProjectForProject(ProjectData project,
...
+      } catch (IOException jiracreateException) 
+      {
+          logger.debug("error creating project with template {}: {}", 
+        		  toBeCreated.projectTemplateKey,
+        		  jiracreateException.getMessage());
+    	  if (jiracreateException.getMessage() != null && jiracreateException.getMessage().startsWith("400")) {
+    		  toBeCreated.projectTypeKey = jiraTemplateType;
+    		  toBeCreated.projectTemplateKey = jiraTemplateKey;
+        	  created = this.callJiraCreateProjectApi(toBeCreated,crowdCookieValue);
+        	  project.projectType = defaultProjectKey;
```
michaelsauter: If I understand this correctly, it creates the project with the default template if the original request fails with a 400. Can you explain *why* we are doing this?

At least we should log on INFO level I'd say.

src/main/java/org/opendevstack/provision/services/JiraAdapter.java
```diff
@@ -285,10 +320,31 @@ protected int createPermissions (ProjectData project, String crowdCookieValue)
...
+    	(s.projectType != null && 
+    		environment.containsProperty(jiratemplateKeyPrefix + s.projectType) && 
+    		projectTemplateKeyNames.contains(s.projectType)) ?
+    		environment.getProperty(jiratemplateKeyPrefix + s.projectType) : jiraTemplateKey;
+    		
+    String templateType =
+    	(s.projectType != null && 
+    		environment.containsProperty(jiratemplateTypePrefix + s.projectType) &&
+    		projectTemplateKeyNames.contains(s.projectType)) ?
+        	environment.getProperty(jiratemplateTypePrefix + s.projectType) : jiraTemplateType;
```
michaelsauter: Those two ternaries are difficult to read for me.

src/main/java/org/opendevstack/provision/controller/ProjectApiController.java
```diff
@@ -68,7 +68,7 @@
   private static final String STR_LOGFILE_KEY = "loggerFileName";
   
   @Autowired
-  private JiraAdapter jiraAdapter;
+  JiraAdapter jiraAdapter;
```
michaelsauter: Why was `private` removed?

Also further down, visibility is mixed between private and package. Can we use one visibility level?

src/main/java/org/opendevstack/provision/services/JiraAdapter.java
clemensutschig: @michaelsauter  - while on confluence you can check for the blueprint - on jira there is NO rest api to make this work, so we have to go kind of trial and error - and HTTP 400 is returned only in this case - so we fall back then to the default one, which we know works. makes sense?

michaelsauter: Yes makes sense. Maybe we can add this as a comment to the code and log on info level to make this more obvious to readers/users.

src/main/java/org/opendevstack/provision/services/JiraAdapter.java
clemensutschig: @michaelsauter  - extracted

src/main/java/org/opendevstack/provision/services/JiraAdapter.java
clemensutschig: @michaelsauter  - -info added now. check the new push.

src/main/java/org/opendevstack/provision/services/JiraAdapter.java
```diff
@@ -108,9 +124,30 @@ public ProjectData createJiraProjectForProject(ProjectData project,
...
+      FullJiraProject created = null;
+      try {
+    	  created = this.callJiraCreateProjectApi(toBeCreated,crowdCookieValue); 
+      } catch (IOException jiracreateException) 
+      {
+          logger.debug("error creating project with template {}: {}", 
+        		  toBeCreated.projectTemplateKey,
+        		  jiracreateException.getMessage());
+    	  if (jiracreateException.getMessage() != null && jiracreateException.getMessage().startsWith("400")) {
+    		  logger.info("Template {} did not work, falling back to default {}",
```
clemensutschig: @michaelsauter  - fyi :)

src/main/java/org/opendevstack/provision/services/ConfluenceAdapter.java
stitakis: For this kind of logic, I sometimes add a public static method because it makes the its logic easy to unit test... it would look like: 
public static String resolveTemplate(String projectTypeKey, String defaultProjectKey, Environment environment
however in this case because the refactored method would have more than 5 params, I would consider to introduce a nested public static helper class maybe called TemplateResolver() and pass environment or other dependencies in the constructor.

clemensutschig: @stitakis  /  @michaelsauter  - I can do what I did in the jiraadapter - to extract it and comment it -good enough?

stitakis: yep, sounds good to me :-)

clemensutschig:
@michaelsauter  / @stitakis  - finally .. I removed all the 0.1 crap on http logic ... enjoy reviewing

src/main/java/org/opendevstack/provision/controller/DefaultController.java
```diff
@@ -86,7 +98,23 @@ String home(Model model) {
...
+    {
+    	if (crowdCookie == null && request != null) {
+    		logger.debug("cookie NULL + " + request);
+    		
+    		Cookie[] cookies = request.getCookies();
+    		for (Cookie cookie : cookies) 
+    		{
+        		logger.debug("cookie" + cookie.getName());
+    			
+    			if (cookie.getName().equals(this.crowdCookieKey))
```
stitakis: Would be useful to add "logger.warn" if this condition is not matched?

src/main/java/org/opendevstack/provision/controller/DefaultController.java
```diff
@@ -86,7 +98,23 @@ String home(Model model) {
...
+    		logger.debug("cookie NULL + " + request);
+    		
+    		Cookie[] cookies = request.getCookies();
+    		for (Cookie cookie : cookies) 
+    		{
+        		logger.debug("cookie" + cookie.getName());
+    			
+    			if (cookie.getName().equals(this.crowdCookieKey)) 
+    			{
+    				crowdCookie = cookie.getValue();
```
stitakis: currently crowdCookie is a method variable which is not assigned to an instance variable or used after this line whithin the method... is this on purpouse? Or the method should fail if the cookie is not set?

clemensutschig:
fixes #85 as well

clemensutschig:
@stitakis  / @michaelsauter  - can you please re-review :)

PR #83, opened 2019-01-25
make crowd sso cookie name configurable (1.0.x)

### LINKED ISSUE #81: Provision app fails with 401 not authenticated in case crowd's sso cookie name was changed - make configurable
On crowd you can change the sso cookie name, like below

image

the codebase assumes `crowd.token_key` to be the standard name - and hence sso will not work if one changed this name in the crowd server config

https://github.com/opendevstack/ods-provisioning-app/blob/master/src/main/java/org/opendevstack/provision/util/CrowdCookieJar.java#L78

clemensutschig:
fixes #81

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,4 @@ bin
 gradle.properties
 /.gradle/
 /docker/app*.jar*
+/src/main/resources/__application.properties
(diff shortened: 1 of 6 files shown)


clemensutschig:
@michaelsauter  / @stitakis  - I found a major glitch .. while this works nice in the unit tests . it fails reality .. :( @Cookievalue (on the method header) - needs a STATIC value name .... any ideas?

PR #99, opened 2019-03-14
add default notification scheme - 1.0.x

### LINKED ISSUE #90: No Default Notification Scheme assigned into Jira projects upon creation
Hi
There is no Default Notification Scheme assigned after a Jira project is created and or the provisioning app is executed, and, has to be added each time manually.

Could be possible to add this feature for mail Notifications for any action on Jira?

clemensutschig:
fixes #90 on 1.0x

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -27,6 +27,9 @@ jira.permission.filepattern=classpath*:permission-templates/jira.permission.*
 
 jira.project.template.key=com.pyxis.greenhopper.jira:gh-scrum-template
 
+# jira notification scheme - attached to the project (10000 is the seed ID for default)
+jira.project.notification.scheme.id=10000
+
 #Confluence properties
 confluence.uri=http://192.168.56.31:8090
 confluence.api.path=/rest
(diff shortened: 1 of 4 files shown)


src/main/resources/application.properties
```diff
@@ -27,6 +27,9 @@ jira.permission.filepattern=classpath*:permission-templates/jira.permission.*
 
 jira.project.template.key=com.pyxis.greenhopper.jira:gh-scrum-template
 
+# jira notification scheme - attached to the project (10000 is the seed ID for default)
+jira.project.notification.scheme.id=10000
```
michaelsauter: Wouldn't this need to go into the `ConfigMap` too? I would probably make this to master only?

clemensutschig: @michaelsauter - this can go to 1.0x as well - it's surgical, and actually also a little bit of a bug .. :)
10000 is the default seed - when you install jira - the chance of ANY reconfig of this is 0 .. that's why I did not include it.

michaelsauter: OK

PR #112, opened 2019-04-11
Implemented: added initial imageLabels key in BC

### LINKED ISSUE #111: Missing Key in BuildConfig after first
```
        {"name":"ods.build.lib.version","value":"production
"}
      ]}]' -n prov-test
Error from server: jsonpatch replace operation does not apply: doc is missing key: /spec/output/imageLabels
```
After first import  of the provisioning app with tailor the pipeline fails because of missing key in BuildConfig

tjaeschke:
An initial imageLabels key is needed on first import of the provisioning app with tailor. If the key is not presents, the build in Jenkins fails. This addition fixes #111


rattermeyer:
Do we have an issue (build fails if tag is missing) that we can link to? Or could you create such an issue?

tjaeschke:
> @tjaeschke this applies to quickstarters aß well, no?

It should. I will extend the docker bc and create a pull request

PR #117, opened 2019-04-17
add kanban board config and new operation to resolve template keys

### LINKED ISSUE #109: Add com.pyxis.greenhopper.jira:gh-kanban-template into configured project types
Today - the only configured (= default) template in the provision app is `com.pyxis.greenhopper.jira:gh-scrum-template`. we should add the kanban one as well

reference:
https://github.com/opendevstack/ods-provisioning-app/blob/master/src/main/resources/application.properties#L33

clemensutschig:
closes #109

### DIFF
--- a/src/main/java/org/opendevstack/provision/services/JiraAdapter.java
+++ b/src/main/java/org/opendevstack/provision/services/JiraAdapter.java
@@ -1,25 +1,9 @@
 package org.opendevstack.provision.services;
 
-import com.atlassian.crowd.integration.springsecurity.user.CrowdUserDetails;
-import com.atlassian.crowd.integration.springsecurity.user.CrowdUserDetailsService;
-import com.atlassian.jira.rest.client.domain.BasicUser;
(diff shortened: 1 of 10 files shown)


clemensutschig:
added the lookup of template key also to confluence adapter - picked up @stitakis sb fix .. (and removed double properties.

src/main/java/org/opendevstack/provision/services/JiraAdapter.java
```diff
@@ -60,11 +51,14 @@
   @Value("${jira.permission.filepattern}")
   private String jiraPermissionFilePattern;
 
+  public static final String jiratemplateKeyPrefix = "jira.project.template.key.";
+  public static final String jiratemplateTypePrefix = "jira.project.template.type.";
```
michaelsauter: I'd keep the casing consistent, so `jiraTemplateKeyPrefix` here and also above `confluenceTemplateKeyPrefix` ...

PR #125, opened 2019-04-25
Travis CI build setup

oalyman:
Travis CI config and gradle build adjustment to be able to build / test on Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,17 @@
+language: java
+install: true
+
+jdk:
+  - oraclejdk8
+
+script:
+  - ./gradlew clean test
+
+before_cache:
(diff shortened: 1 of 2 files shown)


clemensutschig:
Super cool!

clemensutschig:
@oalyman  - can you merge yourself?

.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: java
+install: true
+
+jdk:
+  - oraclejdk8
```
tjaeschke: Just a question. Shouldn't we use openjdk instead of oraclejdk?

oalyman: you're right. Thanks for the hint!

PR #137, opened 2019-05-10
fix security config

oalyman:


clemensutschig:
@oalyman  - where is the sporing 2.14 update ... (which broke this originally .. :)?

PR #165, opened 2019-07-08
Fix for getProject

hugowschneider:
getProject is returning always a project even if the requested project does not exist.

This patch fixes this issue


michaelsauter:
@hugowschneider @gerardcl Sorry to bother, but we need an issue for this. Otherwise it is hard to assemble the changelog items later on. Please open&close an issue and link to here.

gerardcl:
right! done! #174

PR #184, opened 2019-08-01
Bugfix application logs invalid number of existing projects, because …

stefanlack:
…variable 'updatedProject' was not completely initialized


stefanlack:
Excuse me, your right! 

But if the property _updatedProject.repositories_ is not set, further log messages are missleading.
The variable _existingComponentRepos_ is equal to zero:

image

Application logs at line 379 are wrong:
> DEBUG o.o.p.c.ProjectApiController - New quickstarters 1, existing repos: 0, new repos; 1

image

PR #185, opened 2019-08-01
Spotless: Enforce formatting

rattermeyer:
Hi,
just as a discussion: What do you think about enforcing the formatting as Apache Beam does it using spotless:

https://cwiki.apache.org/confluence/display/BEAM/Spotless+pre-commit

BR,
Richard

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,6 +53,7 @@ plugins {
     id 'maven'
     id 'jacoco'
     id 'org.sonarqube' version "2.7"
+    id "com.diffplug.gradle.spotless" version "3.24.0"
 }
 
 apply plugin: 'org.springframework.boot'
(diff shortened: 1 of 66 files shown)


michaelsauter:
I'm fine with this - we do that for our product internally. Not using spotless, but running the google java formatter as part of the pre-commit hook.

stefanlack:
Nice idea.
But: before merging this to master, we should consider that this plugin will change the formatting of 91 files, when the gradle build is executed the first time. It will not only reformat the changed files, but all of them.

For example:
image

@clemensutschig  and @michaelsauter if this ok for you, than we shoul add all reformated files to this pull request, approve it again and merge it to master.

We should to this as soon as possiple, IMHO.

stefanlack:
Reformatted files have been added.

michaelsauter:
@stefanlack Agreed, we should bring all files into the correct formatting.

rattermeyer:
as @stefanlack  has already added the formatted files, we should be fine.
Stefan, can you then approve the PR?

PR #198, opened 2019-09-11
fix #197 invalid implementation of parameter no_nexus

### LINKED ISSUE #197: invalid implementation of parameter no_nexus
The document _README.md_ states

> If you want to build / run locally without NEXUS, you can disable NEXUS by adding the following property to `gradle.properties`:
> 
> ```properties
> no_nexus=true
> ```
> 

Proven to work as expected:
- Setting `no_nexus=true` in the propertis disable nexus repositories --> OK

But the description suggest that one can use `no_nexus=false` to enable the use of nexus repos. This is not the case, since the parameter is read as string value and gradle will evaluate 

`if(no_nexus) {`

to true whenever the variable is defined.

stefanlack:


stefanlack:
Your suggestion is not working, If you set no_nexus=false, the build will use jcenter and mavenCentral and prints:

> using repositories 'jcenter' and 'mavenCentral'

I used following instead
`no_nexus = (project.findProperty('no_nexus') ?: System.getenv('NO_NEXUS') ?: false).toBoolean()`

PR #213, opened 2019-09-16
modify default access for opendevstack users to bitbucket project and…

### LINKED ISSUE #211: modify default access for opendevstack users to bitbucket project and repository
When a bitbucket **project** is created, the following default permission should be granted:
- Every member of the opendevstack-users group (as configured via property `idmanager.group.opendevstack-users`)  should have **read access** at project level

When a bitbucket **repository** is created, the following default permission should be granted:
- Every member of the opendevstack-users group (as configured via property idmanager.group.opendevstack-users) should have **write access** to the generated repository

Any thought about this?

stefanlack:
… repository 

fixes #211

### DIFF
--- a/src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
+++ b/src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
@@ -80,6 +80,9 @@ public class BitbucketAdapter extends BaseServiceAdapter implements ISCMAdapter 
   @Value("${global.keyuser.role.name}")
   private String globalKeyuserRoleName;
 
+  @Value("${idmanager.group.opendevstack-users}")
+  private String openDevStackUsersGroupName;
+
   @Autowired IODSAuthnzAdapter manager;
 
   private static final String PROJECT_PATTERN = "%s%s/projects";
@@ -91,6 +94,11 @@ public class BitbucketAdapter extends BaseServiceAdapter implements ISCMAdapter 
     PROJECT_ADMIN,
     PROJECT_WRITE,
     PROJECT_READ
+  }
+
+  public enum REPOSITORY_PERMISSIONS {
+    REPO_ADMIN,
+    REPO_WRITE
   }
 
   public String createSCMProjectForODSProject(OpenProjectData project) throws IOException {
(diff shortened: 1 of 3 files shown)


stefanlack:
@rattermeyer : thank you for your review. 
@michaelsauter : any thoughts from your site regarding this PR?

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
```diff
@@ -279,8 +289,10 @@ protected RepositoryData callCreateRepoApi(String projectKey, Repository repo)
                   + " - no response from endpoint, please check logs",
               repo.getName(), projectKey));
     }
-    setRepositoryPermissions(data, projectKey, ID_GROUPS, repo.getUserGroup());
-    setRepositoryPermissions(data, projectKey, ID_USERS, technicalUser);
+    setRepositoryAdminPermissions(data, projectKey, ID_GROUPS, repo.getUserGroup());
+    setRepositoryAdminPermissions(data, projectKey, ID_USERS, technicalUser);
+    setRepositoryPermissions(
```
clemensutschig: @stefanlack  - that one I am NOT ok with ... that means everyone has write access .. (and we removed that long time ago ..) why is this needed...?

stefanlack: @clemensutschig because it is our philosophy for OpenDevStack@OPITZ and was requested by @rattermeyer .
But I can see your thoughts about that. We also had some internal discussions.

Suggested Solution: 
- Implementing an property `provision.scm.grantglobalwritetouser` (default value=false). 

By setting to true, the  line `setRepositoryPermissions(data, projectKey, ID_GROUPS, openDevStackUsersGroupName, REPOSITORY_PERMISSIONS.REPO_WRITE);` is executed, otherwise it is not executed


>

stefanlack: Is this ok for you?

clemensutschig: @stefanlack .. there was a rule in the codebase.. if no special perm set.. then its ok... in case of special perm set.. no addtl write rules

rattermeyer:
@clemensutschig can you please check, if you are fine with the changes?

PR #234, opened 2019-09-28
Naming conventions for quickstarters

adyachok:
Introduced logic for  building quickstarter naming convention .

1. Created configuration file with requirements for names selection accordingly to project names supported by different frameworks.
2. Created endpoint for front-end to extract information about specified naming rules.
3. Created filtering logic for processes of creation of a new project as well as update existing one.

### DIFF
--- a/src/main/java/org/opendevstack/provision/controller/ProjectApiController.java
+++ b/src/main/java/org/opendevstack/provision/controller/ProjectApiController.java
@@ -115,7 +115,9 @@ public class ProjectApiController {
     }
 
     // check correctness of a project quickstarters naming, and discard wrongly named
-    newProject.quickstarters = filterQuickstarters(newProject.quickstarters);
+    if (newProject.quickstarters != null) {
+      newProject.quickstarters = filterQuickstarters(newProject.quickstarters);
+    }
 
     if (newProject.specialPermissionSet && !jiraAdapter.isSpecialPermissionSchemeEnabled()) {
       return ResponseEntity.badRequest()
@@ -233,7 +235,7 @@ public class ProjectApiController {
       // Find right name for a component
       for(Job job: availableQuickstarters) {
         if(job.getId().equals(component_type)) {
-          component_name = job.getId();
+          component_name = job.getName();
           break;
         }
       }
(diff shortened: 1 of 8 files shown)


gerardcl:
LGTM! I will test it and let's have it into 1.2.x too!

adyachok:
Sure. I will add tests.
I think we need to fix FE part as well. I can include regex for FE part in a rules object, however I would like to invite a front-end developer for an increment in FE part.

clemensutschig:
1.2 i guess.. ;-)

gerardcl:
hi @adyachok and all, is this ok if we do not put this feature yet in ODS1.2.x initial release but later on? I want to assure we are not adding any special change at this moment, does it make sense?

stefanlack:
Created https://github.com/adyachok/ods-provisioning-app/pull/1

adyachok:
@stefanlack I nice work.

clemensutschig:
@stefanlack - go for 2.0?

stefanlack:
@clemensutschig I'm afraid but I do not understand your last comment. I have created an PR on Oct.7, @adyachok answered with "nice work". So @adyachok  and you may merge my pr into #234  and finish this.

michaelsauter:
I propose to move this to v3.

stefanlack:
I also propose to move this to v3

src/main/java/org/opendevstack/provision/controller/ProjectApiController.java
```diff
@@ -209,6 +217,58 @@
...
   }
 
+  private List<Map<String, String>> filterQuickstarters(List<Map<String, String>> quickstarters) {
+    /**
+     * Pushes every quckstarter object though chain of filters.
+     *
+     * @param quickstarters - selected by user quickstarters to create
+     * @return list of correctly named parameters
+     */
+    List<Job> availableQuickstarters = rundeckAdapter.getQuickstarters();
```
stitakis, line 227: rundeck is deprecated and not supported anymore

src/test/java/org/opendevstack/provision/controller/ProjectApiControllerTest.java
```diff
@@ -108,9 +125,14 @@ private void initOpenProjectData() {
     data.description = "Description";
 
     Map<String, String> someQuickstarter = new HashMap<>();
-    someQuickstarter.put("key", "value");
+    someQuickstarter.put("component_id", "fe-angular-7");
```
stitakis, line 128: numbers are not supported anymore in component id

stitakis:
I haven't taken a deep look into this PR, but first impression is that needs to be "adapter" to latest version from master.

michaelsauter:
Let's close this. The problem this PR addresses is fixed by other means - without the extra complexity. Let's come back to it when we have users needing this.

PR #238, opened 2019-10-01
Detect Firefox

### LINKED ISSUE #110: provisioning app does not support Firefox, confused as if it was IE
Screenshot 2019-04-08 at 16 33 09

FelixWoRe:
Detect Firefox separated from Internet Explorer.

fixed #110


FelixWoRe:
@michaelsauter @ManuelFeller

PR #241, opened 2019-10-02
invalid property in application-oauth2.properties

stefanlack:
fix property spring.security.oauth2.client.provider.keycloak.authorization-uri


clemensutschig:
@stefanlack 1.2? If so Please merge

PR #250, opened 2019-10-09
add simple manual test that can be used to clean up  atlassian resources

### LINKED ISSUE #249: provide a cleanup util for atlassian projects
Feature request: provide a utiliy class that can be used to clean up jira projects, confluence spaces and bitbucket projects that are created during developer testing.

Implement a test class that can be executed manually. The class should delete the resources via call to the  rest api.

The needed credentials and urls should be taken from a spring boot profile.

**Describe alternatives you've considered**
None.

stefanlack:
closes #249

### DIFF
--- a/src/main/java/org/opendevstack/provision/util/rest/RestClientCall.java
+++ b/src/main/java/org/opendevstack/provision/util/rest/RestClientCall.java
@@ -170,8 +170,8 @@ public class RestClientCall {
     return this;
   }
 
-  public RestClientCall url(String urlFormat, Object ... args) {
-    return url(String.format(urlFormat,args));
+  public RestClientCall url(String urlFormat, Object... args) {
+    return url(String.format(urlFormat, args));
   }
 
   public RestClientCall credentials(CredentialsInfo credentialsInfo) {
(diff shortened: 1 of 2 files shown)


stefanlack:
Alternatives:
1. move  from test-class to separate main class, But: does not change the fact, that user has to modify source code in order to clean up

2. Move clean up implementation to service class, add new rest api for clean up functionality. That should be protected somehow, because potential damage can be very high if functionality is missused. Therefore, we have to introduce a new opendevstack role, e.g. opendevstack-superadministrators, only users in this role/group have rights to execute this functionality

3. Add cleanup functionality to provisioning app gui. In this case, we need the opendevstack-superadministrators group as well.

Evaluation
- 1. does not solve the issue, that one has to change code in order to cleanup bitbucket projects
- 2. and 3. are much nicer, but introduce some work.

Consideration: do we reallyshould invest more effort, just for providing some functionality that is only used from a developer in some rare cases?

stefanlack:
@clemensutschig @michaelsauter @rattermeyer may you please give an opinion to this?

PR #265, opened 2019-11-06
Replace Rundeck with (Jenkins) pipelines #283

kiwo:
Addresses https://github.com/opendevstack/ods-project-quickstarters/issues/283.


Jenkinsfile
```diff
@@ -38,12 +38,14 @@ def stageBuild(def context) {
...
   stage('Build') {
+    sh 'echo ${APP_DNS}'
+    sh 'openssl s_client -showcerts -connect ${APP_DNS}:443 < /dev/null | openssl x509 -outform DER > docker/derp.der'
     withEnv(["TAGVERSION=${context.tagversion}", "NEXUS_USERNAME=${context.nexusUsername}", "NEXUS_PASSWORD=${context.nexusPassword}", "NEXUS_HOST=${context.nexusHost}", "JAVA_OPTS=${javaOpts}","GRADLE_TEST_OPTS=${gradleTestOpts}","ENVIRONMENT=${springBootEnv}"]) {
       def status = sh(script: "./gradlew clean build --stacktrace --no-daemon", returnStatus: true)
       junit 'build/test-results/test/*.xml'
       if (status != 0) {
         error "Build failed!"
-      }
+      }
```
michaelsauter: Please remove the whitespace

michaelsauter:
@kiwo Maybe we can hold off merging that a little so that the time window is shorter between merging this and having all the quickstarters available as Jenkins jobs ... what do you think?

src/main/java/org/opendevstack/provision/model/rundeck/Job.java
```diff
@@ -16,18 +16,38 @@
```
michaelsauter: Should this be moved to another package? `rundeck.Job` looks weird when there is no Rundeck at all :)

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
```diff
@@ -235,73 +230,86 @@ public OpenProjectData createPlatformProjects(OpenProjectData project) throws IO
...
 
-    Job job = jobStore.getJobByNameOrId(jobNameOrId);
+    String url;
+    Execution execution = new Execution();
 
-    String url = String.format("%s%s/job/%s/run", rundeckUri, rundeckApiPath, job.getId());
+    String projID = Objects.toString(options.get("PROJECT_ID"));
 
-    Execution execution = new Execution();
+    if (projectCreateOpenshiftJob.equals(jobNameOrId)) { // we create something
```
michaelsauter: The comment is not clear to me.

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
```diff
@@ -235,73 +230,86 @@ public OpenProjectData createPlatformProjects(OpenProjectData project) throws IO
...
+              + "&jenkinsfile_path="
+              + job.jenkinsfilePath
+              + "&component=ods-quickstarter-"
+              + job.getName()
+              + "-"
+              + component_id;
+      execution.branch = job.branch;
+      execution.repository = job.gitURL;
+    }
     if (options != null) {
```
michaelsauter: Took me a long time to realise that HERE all the magic happens (e.g. setting the project to create). Also, it seems weird that we already use `options` above, assuming it is non-null and then have this check here ...

Can we make the flow clearer?

src/main/resources/application.properties
```diff
@@ -98,6 +98,8 @@ rundeck.project.group.quickstarter=quickstarts
 rundeck.project.group.openshift=openshift
 rundeck.project.openshift.create.name=create-projects
 
+jenkinspipeline.quickstarter={beSpringBoot: 'http://192.168.56.31:7990/scm/prov/quickstarter-jenkins.git#master/beSpringBoot/Jenkinsfile'}
```
clemensutschig: why do we have to encode an IP here .. we have bitbucket host etc already in the property file?!

src/main/resources/application.properties
```diff
@@ -98,6 +98,8 @@ rundeck.project.group.quickstarter=quickstarts
...
+jenkinspipeline.create-ods-projects-job=http://192.168.56.31:7990/scm/opendevstack/ods-core.git#production/create-projects/Jenkinsfile
```
clemensutschig: see above

michaelsauter: Yes, absolutely! This was just temporary so that we can discuss the current state. @kiwo I think it would be good to not use an absolute URI path at all but to narrow the "scope" to a repository underneath `opendevstack` ... so either users pick e.g. `ods-core` or `ods-quickstarters` ... or their own `my-super-cool-quickstarters`, but for security reasons (and access rights) it is probably good to restrict to `opendevstack` project on your BitBucket instance.

src/main/resources/application.properties
clemensutschig: The default should be the Bitbucket repo that is alread defined ...

src/main/resources/application.properties
rattermeyer: Depends when you can change the quickstarter URL. If you could specifiy alternative repositories at runtime, then I would completely agree on your comment. If it is something that is once set at deployment time, we could leave the complete URL in.

michaelsauter: @rattermeyer Not quite, because it affects the webhook proxy. If you have full URLs that you send to the proxy/Jenkins than it is problematic if someone would be able to bypass security. better to just be able to build pre-configured projects. See https://github.com/opendevstack/ods-provisioning-app/issues/270.

src/main/resources/application.properties
michaelsauter: As stated in https://github.com/opendevstack/ods-provisioning-app/issues/270, I want the whole host to go away - we don't need it.

clemensutschig:
@kiwo  - what's the status of this?!

kiwo:
merging is delayed until the new jenkins quickstarters are completed and this pr is using the new format described in #270

Issue #261, opened 2019-10-31
Connecting to rundeck to fetch jobs .. fails with "closed"


clemensutschig:
**Describe the bug**

Whitelabel Error Page
This application has no explicit mapping for /error, so you are seeing this as a fallback.

Thu Oct 31 16:00:16 GMT 2019
There was an unexpected error (type=Internal Server Error, status=500).
closed

**To Reproduce**
provide wrong rundeck authentication means / token

**Expected behavior**
A clear error message coming from the server, containing the cause!

**Affected version (please complete the following information):**
 - OpenDevStack [1.2.0]

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
```
  | 2019-10-31 16:00:16 DEBUG o.s.s.w.c.SecurityContextPersistenceFilter - SecurityContextHolder now cleared, as request processing completed
  | 2019-10-31 16:00:16 ERROR o.a.c.c.C.[.[.[.[dispatcherServlet] - Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is java.lang.IllegalStateException: closed] with root cause
  **| java.lang.IllegalStateException: closed**
  | at okio.RealBufferedSource.rangeEquals(RealBufferedSource.java:398)
  | at okio.RealBufferedSource.rangeEquals(RealBufferedSource.java:392)
  | at okhttp3.internal.Util.bomAwareCharset(Util.java:431)
  | at okhttp3.ResponseBody.string(ResponseBody.java:174)
  | at org.opendevstack.provision.util.rest.RestClient.sendPreAuthRequest(RestClient.java:83)
  | at org.opendevstack.provision.util.rest.RestClient.execute(RestClient.java:43)
  | at org.opendevstack.provision.services.RundeckAdapter.getJobsFromServer(RundeckAdapter.java:254)
  | at org.opendevstack.provision.services.RundeckAdapter.initializeJobsForGroup(RundeckAdapter.java:152)
  | at org.opendevstack.provision.services.RundeckAdapter.getQuickstarters(RundeckAdapter.java:116)
  | at org.opendevstack.provision.controller.DefaultController.provisionProject(DefaultController.java:98)
  | at sun.reflect.GeneratedMethodAccessor101.invoke(Unknown Source)
```

**Additional context**

Problem is the double string read ... :(
[...]

Issue #264, opened 2019-11-06
An error occured while creating project XXXX, reason closed - cleanup attempted, but [{BUGTRACKER_PROJECT=1}] components are still there!


segator:
**Describe the bug**
When creating projects on prov-app using API, randomly we get this kind of error

**To Reproduce**
prov-app api project creation with ODS.

**Expected behavior**
Project created succesffully.

**Affected version**
 - OpenDevStack  1.2.x

**Log Output**
```
10-30 16:52:30 loggerFileName=US056 [http-nio-8080-exec-12] DEBUG o.o.provision.storage.LocalStorage - Project BIALLCYTE is in std format
2019-10-30 16:52:30 loggerFileName=US056 [http-nio-8080-exec-12] DEBUG o.o.provision.storage.LocalStorage - File /opt/provision/history/20191030041744-STARRSEQ.txt:
2019-10-30 16:52:30 loggerFileName=US056 [http-nio-8080-exec-12] DEBUG o.o.provision.storage.LocalStorage - Project STARRSEQ is in std format
2019-10-30 16:52:30 loggerFileName=US056 [http-nio-8080-exec-12] DEBUG o.o.provision.storage.LocalStorage - Could not find project with id US056
2019-10-30 16:52:30 loggerFileName=US056 [http-nio-8080-exec-12] DEBUG o.o.p.services.StorageAdapter - Project US056 not affected from cleanup, as it was never stored
2019-10-30 16:52:30 loggerFileName=US056 [http-nio-8080-exec-12] DEBUG o.o.p.c.ProjectApiController - Overall cleanup status of project: {BUGTRACKER_PROJECT=1} components left
2019-10-30 16:52:30 loggerFileName=US056 [http-nio-8080-exec-12] ERROR o.o.p.c.ProjectApiController - An error occured while creating project US056, reason closed - cleanup attempted, but [{BUGTRACKER_PROJECT=1}] components are still there!
java.lang.IllegalStateException: closed
        at okio.RealBufferedSource.rangeEquals(RealBufferedSource.java:398)
        at okio.RealBufferedSource.rangeEquals(RealBufferedSource.java:392)
        at okhttp3.internal.Util.bomAwareCharset(Util.java:431)
        at okhttp3.ResponseBody.string(ResponseBody.java:174)
        at org.opendevstack.provision.util.rest.RestClient.sendPreAuthRequest(RestClient.java:83)
[...]

PR #266, opened 2019-11-07
bugfix #264

stefanlack:
Create preAuthContent map for every request. (Previous code was valid if technical user is used, but for crowd based authentication the user and password may be changing for differnt calls of this method)


clemensutschig:
fixes #261 and #264

PR #271, opened 2019-11-15
Adjust configuration to new config sample

michaelsauter:
Uses the new, unified files and adds PROV_APP_ prefix to the params.

Relates to https://github.com/opendevstack/ods-core/issues/219.

I left the Rundeck stuff in on purpose ... just want to make it "work", we can kill all the Rundeck things later.

### DIFF
--- a/ocp-config/prov-app/cm.yml
+++ b/ocp-config/prov-app/cm.yml
@@ -14,8 +14,8 @@ objects:
       logging.level.org.springframework=INFO
       logging.level.org.springframework.security=DEBUG
       logging.level.com.atlassian=INFO
-      logging.level.com.atlassian.crowd=${LOG_LEVEL_ATLASSIAN_CROWD}
-      logging.level.org.opendevstack=${LOG_LEVEL_OPENDEVSTACK}
+      logging.level.com.atlassian.crowd=${PROV_APP_LOG_LEVEL_ATLASSIAN_CROWD}
+      logging.level.org.opendevstack=${PROV_APP_LOG_LEVEL_OPENDEVSTACK}
 
       # log file
       logging.file=provision.log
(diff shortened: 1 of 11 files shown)


ocp-config/prov-app/cm.yml
```diff
@@ -131,24 +131,24 @@ objects:
...
-- name: JIRA_URI
+- name: PROV_APP_JIRA_URI
   required: true
   description: the URI of JIRA used to create new jira projects  
-- name: CONFLUENCE_URI
+- name: PROV_APP_CONFLUENCE_URI
   required: true
   description: the URI of confluence used to create new confluence spaces
-- name: BITBUCKET_URI
+- name: PROV_APP_BITBUCKET_URI
```
clemensutschig: @michaelsauter bitbucket URI is used in many placed, or? .. if so - this should not be "prov_app" :)

michaelsauter: Yes it is. BUT I was not sure if I should use the "globally configured" one here as we have two instances (prov-dev and prov-test) and one might want to configure them differently?

IMHO eventually ODS should ship with one central ProvApp instance in the `cd` namespace (meaning to treat it similar to Nexus/SonarQube). If we were to do that for e.g. v3/v4 then I would propose to just use the globally configured params then. For now though I'd stick with separate configs.

Do you agree or would you rather have a fixed single Atlassian stack for both prov-dev and prov-test instances?

clemensutschig: I would use the global one ... otherwise - you need to introduce a double pair (dev / test :))

tjaeschke: I'm with @michaelsauter.

clemensutschig: ok  then we go with it :)

PR #288, opened 2019-11-29
bugfix #280 Quickstarters are missing a description

stefanlack:
fxes #280

PR #289, opened 2019-11-29
Configure ods image and ref

michaelsauter:
There is some overlap now with the configuration of the quickstarters, but I don't want to change that before @stefanlack your changes with the description are in.

This change allows to configure the ref/image to use e.g. in the quickstarter Jenkinsfiles, which makes it quite easy to change out new stuff :)

### DIFF
--- a/ocp-config/prov-app/cm.yml
+++ b/ocp-config/prov-app/cm.yml
@@ -25,7 +25,7 @@ parameters:
   description: the API host of the OCP cluster
 - name: OPENSHIFT_CONSOLE_HOST
   required: true
-  description: the console host of the OCP cluster 
+  description: the console host of the OCP cluster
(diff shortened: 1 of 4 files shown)


stefanlack:
@michaelsauter thanks for waiting with this. My PR https://github.com/opendevstack/ods-provisioning-app/pull/288 is ready for review.

michaelsauter:
@stefanlack I'd still would like to merge this independently. If we make more adjustments, those would be a separate PR.

I think eventually, in the quickstarter config we should take apart the `url` field into more concrete fields (e.g. `repo`, `jenkinsfile` and optionally a `git-ref`, which if not set, defaults to the global `ods.git-ref` as set in the PR here). But that is a bigger issue which I'm not sure will make the ODS 2 cut.

PR #314, opened 2019-12-06
ds-dash quickstarter

gerardcl:
relates to https://github.com/opendevstack/ods-quickstarters/issues/77

### DIFF
--- a/ocp-config/prov-app/cm.yml
+++ b/ocp-config/prov-app/cm.yml
@@ -92,7 +92,7 @@ objects:
       server.connection-timeout=90000
       spring.session.timeout=240
       spring.mvc.async.request-timeout=90000
-      
+
       # server port
       server.port=8080
       server.use-forward-headers=true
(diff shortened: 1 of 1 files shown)


clemensutschig:
@gerardcl  - can you please merge this into master?

michaelsauter:
No, this should be closed. Dash is not a quickstarter in the official repo - please see my comment. 

Let's see actual usage, and until we know for sure that we want to support and maintain this under the open source umbrella, we should not pick it up. **We need fewer, not more quickstarters as part of the official distribution**.

gerardcl:
AFAIK it is already being used by many different DS colleagues, @sklingel correcte me if I am wrong, but they can anyway use it from jupyter QS skeleton or internal QS too. I am fine on closing it.

michaelsauter:
Let's close this for now as the other repo does not contain the quickstarter. 

Next step is to develop a nice way to configure quickstarters (that will be done for v3) and then we decide which quickstarter goes into which repo for v3 :)

PR #316, opened 2019-12-09
improve wording

henrjk:
I would like to see this wording change (see PR #312).


henrjk:
I updated the messages, this looks better to me as well.

PR #320, opened 2019-12-12
Bugfix gh 318 session timeout issues

### LINKED ISSUE #318: Session timeout issues
**Describe the bug**
There is an issue with outgoing request (e.g. to Jira) not being authenticated properly as the requests use memory scoped to the Spring session. When that happens, the user is not redirect to the login, but an exception is thrown.

**To Reproduce**
1. Login to prov-app
2. Wait for 5 mins
3. Do something that triggers external call (e.g. change project ID when creating a new project)

**Expected behavior**
User should be sent to login.

**Affected version (please complete the following information):**
 - OpenShift: all
 - OpenDevStack: 1.2.x and up

FYI @henrjk @clemensutschig @oalyman Just capturing this here on Github.

stitakis:
fixes #318


Jenkinsfile
```diff
@@ -15,7 +15,7 @@ library identifier: 'ods-library@production', retriever: modernSCM(
 
 // See readme of shared library for usage and customization.
 odsPipeline(
-  image: "${dockerRegistry}/cd/jenkins-slave-maven",
+  image: "${dockerRegistry}/cd/jenkins-slave-maven:1.2.x",
```
clemensutschig: @michaelsauter  / @stitakis  _ do we tag now on master?

src/main/java/org/opendevstack/provision/config/ControllerAdviceExceptionHandler.java
```diff
@@ -0,0 +1,44 @@
...
+  @Autowired private IODSAuthnzAdapter manager;
+
+  @ExceptionHandler(MissingCredentialsInfoException.class)
+  public ResponseEntity handleException(MissingCredentialsInfoException ex, WebRequest request) {
+
+    String username = request.getRemoteUser();
+
+    // Clean up spring security context
+    SecurityContextHolder.clearContext();
+    logger.info("Spring security context cleared!!");
```
clemensutschig: debug?

src/main/java/org/opendevstack/provision/config/ControllerAdviceExceptionHandler.java
```diff
@@ -0,0 +1,44 @@
...
+
+    try {
+      // Invalidate indentity
+      manager.invalidateIdentity();
+      logger.info("Identity invalidated!");
+    } catch (Exception e) {
+      logger.warn("Error while trying to invalidate identity!");
+    }
+
+    logger.error("Removed user session for user '" + username + "'!", ex);
```
clemensutschig: see above

src/main/java/org/opendevstack/provision/services/BaseServiceAdapter.java
```diff
@@ -87,12 +88,16 @@ public RestClientCall httpDelete() {
   private RestClientCall authenticatedCall(HttpVerb verb) {
     RestClientCall call = notAuthenticatedCall(verb);
 
-    if (useTechnicalUser) {
-      return call.basicAuthenticated(new CredentialsInfo(userName, userPassword));
+    try {
+      if (useTechnicalUser) {
+        return call.basicAuthenticated(new CredentialsInfo(userName, userPassword));
+      }
+      CredentialsInfo credentialsInfo =
```
clemensutschig: the tech user thing should never be null if in use - and they are configured hard..
I woudl move the try to l95

PR #323, opened 2019-12-13
Show cd user comment on creation only

michaelsauter:


src/main/resources/static/js/client.js
```diff
@@ -448,7 +448,11 @@ function summarize(data) {
   }
 
   if (data.cdUser != null && data.cdUser.length > 0) {
-    $("#dataCdUser").html('Project specific CD user<br/>After project creation is complete, you will have to set the <a href="' + data.platformCdEnvironmentUrl + '/browse/secrets/cd-user-with-password" target="_blank">cd-user-with-password secret</a> in the OpenShift webconsole. The actual password can be set in base64 encoding via "Actions > Edit YAML".');
+    var cdUserHtml = 'Project specific CD user';
+    if (actionType != 'updated') {
```
henrjk: I would prefer this to check for === 'created'

michaelsauter: Yes you are right, that would be better. My JavaScript skills are a bit rusty these days :)

Given that a complete overhaul is in order I'll still go ahead with this as-is for ODS 2.

PR #339, opened 2020-01-28
Replace Travis CI build with github actions

oalyman:
As we had problems with Travis CI previously when creating the integration tests for the ods-quickstarters we switched to github actions (https://github.com/opendevstack/ods-quickstarters/blob/master/.github/workflows/continuous-integration-workflow.yml).

This change switches form Travis CI to a github actions workflow for building the provisioning app.


oalyman:
@clemensutschig For now, I just wanted to align it technically with using github actions.
Having integration tests inside openshift will involve a lot more effort, as one also needs to take care of running the whole atlassian stack etc. or mock it somehow.
So I would consider that a different issue/improvement.

clemensutschig:
@oalyman  - I strongly suggest to use the new piece, to see if the "embedded" build works .. can you try?

michaelsauter:
I'm with @oalyman here. I would also keep the tests-as is for now. What @hugowschneider is great, but I think there is value in being able to test the prov-app in isolation, and those tests should run here on Github as well - unmodified.

As @oalyman suggested, we could create another issue that is about testing the pro-app within the cluster context, based on @hugowschneider work ...

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,25 @@
...
+    - name: Build with Gradle
+      run: ./gradlew clean spotlessCheck test
+      env:
+        NO_NEXUS: true
+    - uses: actions/cache@v1
+      with:
+        path: ~/.gradle/caches
+        key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle') }}
+        restore-keys: |
+          ${{ runner.os }}-gradle-
```
michaelsauter: EOF newlines, please :)

clemensutschig:
@oalyman @michaelsauter  . ok .. sold :)

michaelsauter:
@oalyman should there be a build for this PR already? Or does this only works once in master?

PR #372, opened 2020-03-06
adds option to disable logout from identity manager

### LINKED ISSUE #365: Logout from identity manager should be optional
Currently the OAuth2 configuration requires to configure the logout identity manager endpoint.
This is not optional.
Problem is that in some environment users/admins do not want to logout from identity manager but just from the prov-app.

Have a configuration property that allow to disable the logout redirect to identity manager for deployments where users do not want to logout from identity manager but just from prov-app.
Property could be named: ```idmanager.disable-logout-redirect``` and should default to ```false```

stitakis:
backport to ODS 2.x fixes #365


stitakis:
@michaelsauter @oalyman @metmajer @segator please have a quick look at this... the corresponding change in master branch #370 was approved and merged already

PR #383, opened 2020-03-16
adds json response for pre condition checks backport 2.x

### LINKED ISSUE #330: prov-app preflight check
**Check Before create project**
* Return an error code/message  on the API when jira user used to create project doesn't have apropiate permissions to  create/update/rollback(delete) project
* Return an error code/message  on the API when jira user/groups injected in the API doesn't exists. This case should return an especial code because we can implement a retry in the system that call prov-app API, because in our case groups are created in the same pipeline but sincronized to jira in async way, 

This should apply to jira, confluence and bitbucket.

stitakis:
partially fixes #330


stitakis:
closing, we decided today to not include the pre flight checks in 2.x

stitakis:
closing, we decided today to not include the pre flight checks in 2.x

PR #385, opened 2020-03-16
adds default value for logback file path backport 2.x

### LINKED ISSUE #384: logback configuration does not default to old hardcoded file path
**Describe the bug**
logback configuration does not default to old hardcoded file path

**Expected behavior**
for backwards compatibility logback configuration should default to all hardcoded file path

stitakis:
Fixes #384 
Adds backward compatibility for logback configuration. File path defaults to old value.

### DIFF
--- a/ocp-config/prov-app/cm.yml
+++ b/ocp-config/prov-app/cm.yml
@@ -67,9 +67,6 @@ parameters:
 - name: PROV_APP_LOG_LEVEL_OPENDEVSTACK
   required: true
   description: Log level of OpenDevStack package
-- name: LOGGING_FILE_PATH
-  required: true
-  description: File system location for the log files
 objects:
 - kind: ConfigMap
   metadata:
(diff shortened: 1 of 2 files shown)


stitakis:
@michaelsauter not sure if we need this PR because in ```.../ocp-config/prov-app/cm.yml``` the value of property ```LOGGING_FILE_PATH``` is equals ```true```(https://github.com/opendevstack/ods-provisioning-app/blob/df2fcc0e6aa612c07849fdf839cb31385cd15d4a/ocp-config/prov-app/cm.yml#L70). The deployment of ProvApp using tailor will fail is the given environment configuration does not include the property ```LOGGING_FILE_PATH```. Am I right? 
If I'm right, to make ```2.x``` backwards compatible, we need:
A. this PR 
B. and additionally we need to remove the property ```LOGGING_FILE_PATH``` from the ```.../ocp-config/prov-app/cm.yml``` file or flag it as optional.

michaelsauter:
@stitakis correct. you can provide a default value in the CM

stitakis:
@michaelsauter I reconsider the solution... because a default value is already given in the `logback-spring.xml` configuration I'll set the `.../ocp-config/prov-app/cm.yml` in definition of property `LOGGING_FILE_PATH` from `required: true` to `required: false`. This way we provide a default value and a way to overwrite it. Sounds good?

stitakis:
I finally removed the parameter `LOGGING_FILE_PATH`. Having it as not required didn't seem to me to add a value. My local test worked fine so far. This is ready to go be merged

PR #394, opened 2020-03-18
add configuration for webhook creation per quickstarter

### LINKED ISSUE #392: Provide a configuration to blacklist webhook creation by component type
Sometimes a webhook for your quickstarter is not desired. E.g. the release-manager quickstarter should not create builds while you configure / adapt the configuration of components that are part of your release.

Be able to configure a blacklist of component types that avoids webhook creation for those.

**Describe alternatives you've considered**
none

**Additional context**
none

oalyman:
Resolves #392

### DIFF
--- a/src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
+++ b/src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
@@ -25,11 +25,13 @@ import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import javax.annotation.PostConstruct;
 import org.apache.commons.lang3.NotImplementedException;
 import org.opendevstack.provision.adapter.IODSAuthnzAdapter;
 import org.opendevstack.provision.adapter.ISCMAdapter;
 import org.opendevstack.provision.adapter.exception.AdapterException;
 import org.opendevstack.provision.adapter.exception.CreateProjectPreconditionException;
+import org.opendevstack.provision.config.JenkinsPipelineProperties;
 import org.opendevstack.provision.model.OpenProjectData;
 import org.opendevstack.provision.model.bitbucket.BitbucketProject;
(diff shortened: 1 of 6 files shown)


src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
```diff
@@ -430,7 +448,17 @@ public static boolean containsGroup(JsonNode json, String group) {
   }
 
   // Create webhook for CI (using webhook proxy)
-  protected void createWebHooksForRepository(RepositoryData repo, OpenProjectData project) {
+  protected void createWebHooksForRepository(
+      RepositoryData repo, OpenProjectData project, String componentType) {
+
+    if (noWebhookComponents.contains(componentType)) {
```
stitakis: sharing with you my thought here... I'm not sure if this place is a good place for this `if (noWebhookComponents.contains(componentType))` condition or I would be better to place it in the caller function. Apart from this the implementation looks good to me. It is up to you if you prefer to move it to the caller function.

src/main/resources/application.properties
```diff
@@ -185,6 +186,7 @@ jenkinspipeline.quickstarter.e2e-cypress.repo=ods-quickstarters
 
 jenkinspipeline.quickstarter.release-manager.desc=Other - Releasemanager
 jenkinspipeline.quickstarter.release-manager.repo=ods-quickstarters
+jenkinspipeline.quickstarter.release-manager.create-webhook=false
```
stitakis: @oalyman this line was not included in the `cm.yml`... have you forgot it?

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
oalyman: Yes, I have chosen it this way because with the current implementation it was easier to test this way. Otherwise I would need to set up a mock party with a lot of interactions.

PR #395, opened 2020-03-18
add configuration for webhook creation per quickstarter 2.x backport

oalyman:
backport for 2.x for issue #392


src/main/java/org/opendevstack/provision/config/Quickstarter.java
```diff
@@ -47,7 +51,7 @@ public static Quickstarter adminjobQuickstarter(
       String desc,
       Optional<String> branch,
       Optional<String> jenkinsfile) {
-    Quickstarter result = new Quickstarter(name, repo, desc, branch, jenkinsfile);
+    Quickstarter result = new Quickstarter(name, repo, desc, branch, jenkinsfile, false);
```
michaelsauter: Isn't the default supposed to be `true`? Or am I confusing things now?

PR #398, opened 2020-03-27
move provision app to global shared library

clemensutschig:
This is the uptake of global shared lib import for prov app & also the one part to add jenkins file bc for the mro shared lib.. 

@michaelsauter - should we extend to also see for docgen? - alternative, seed via tailor in the respective component repos .. (just let me know :))


stitakis:
@clemensutschig changes look good to me. I'll test it before approving if integration test still needs to be completed.

Jenkinsfile
```diff
@@ -1,19 +1,14 @@
...
-  sharedLibraryRepository = env.SHARED_LIBRARY_REPOSITORY
+  odsImageTag = env.ODS_IMAGE_TAG ?: 'latest'
   dockerRegistry = env.DOCKER_REGISTRY
 }
 
-library identifier: "ods-library@${odsGitRef}", retriever: modernSCM(
-  [$class: 'GitSCMSource',
-   remote: sharedLibraryRepository,
-   credentialsId: credentialsId])
+@Library('ods-jenkins-shared-library@2.x') _
```
michaelsauter: Why is this not using the value from `env.ODS_GIT_REF`? Is this what you just told me about the params issue? Or am I confused now? It would be great not to have specific versions here to avoid merge conflicts and having stale versions because we forget to update ...

clemensutschig: @michaelsauter  - this is what i mean -NOT working .. I tried all sorts of things .. and this does not work :(

michaelsauter: Ok got it. I would like to have some time this week to experiment a little. Maybe there is a way to get it to work.

clemensutschig: enjoy (I tried with params / properties / .... :))

michaelsauter: Okay I gave this a spin. It is, as you figured out, not possible with `@Library`. However, it is possible using `library`. See https://jenkins.io/doc/book/pipeline/shared-libraries/#loading-libraries-dynamically. I tried that and it worked.

From what I understand, the main difference is:

> Whereas the @Library annotation prepares the “classpath” of the script prior to compilation, by the time a library step is encountered the script has already been compiled. Therefore you cannot import or otherwise “statically” refer to types from the library.

Which simply means that one cannot use classes from the `src/` directory well.

I want to give this some more thought, but I think I am OK with that restriction:
- we expose the main building blocks under `vars/` anyway
- people can still change to the `@Library` annotation if needed, and then just hardcode the version they need

Thoughts?

clemensutschig: so - no real drawbacks then? ... what snippet did you make fly?

michaelsauter: I made up my mind. My recommendation is this:

* Use `@Library('ods-jenkins-shared-library@@ods_git_ref@') _` in the `Jenkinsfile.template` of quickstarters, which will render to e.g. `@Library('ods-jenkins-shared-library@2.x') _`
* Use `library("ods-jenkins-shared-library@${odsGitRef}")` in files like here where we want to dynamically resolve. This style would be used for the `Jenkinsfile` (the provisioning one) in the quickstarters

clemensutschig: @michaelsauter  - you rock! ... changed and pushed :)

segator:
ods-library alias doesn't exist anymore in a clean jenkins (ODS 2.x) so build a prov-app in a clean environment fails, can you cherry pick this change to 2.x? this will fix the issue. (already tested in our dev environment)

clemensutschig:
@segator  - just send a pr

PR #406, opened 2020-04-06
change certificate authority configuration to runtime instead of imag…

oalyman:
This PR changes the way the prov-app configures the certificate authority.

Before the CA was baked into the image during image build time. This limited the flexibility (building locally, etc.) and was a blocker for publishing the container image to the public docker repo.

Now if not configure otherwise the default Openshift service-ca will be used. See https://docs.openshift.com/container-platform/3.11/dev_guide/secrets.html#service-serving-certificate-secrets.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,3 +39,5 @@ gradle.properties
 /src/main/resources/application-testods.properties
 
 /src/main/resources/history-local/
+
+.DS_Store
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -37,8 +37,6 @@ def stageBuild(def context) {
     springBootEnv = 'dev'
   }
   stage('Build') {
-    sh 'echo ${APP_DNS}'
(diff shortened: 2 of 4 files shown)


oalyman:
@clemensutschig No. What would you wan't to change?

michaelsauter:
@clemensutschig This is the first of a series of PRs. Next steps are:

* build image in Github and push image to DockerHub
* create templates in `ods-core` for prov-app
* clean up templates in `ods-provisioning-app`

@oalyman did you check whether it is possible to restrict permissions again in the entrypoint script after the cert has been written? That would be cool.

stitakis:
@michaelsauter which "create templates" are you referring to?

oalyman:
@michaelsauter Just for the record. Unfortunately it is not possible to chmod the cacerts file back to readonly by root at runtime.
But I will make the chmod more narrow to only target the cacerts file.

stitakis:
It looks good to me!

PR #409, opened 2020-04-07
includes git server url in webhook proxy build payload on create proj…

### LINKED ISSUE #407: Jenkins create project job uses env.BITBUCKET_HOST and not configured application.properties' one
**Describe the bug**
If one wants to use provision app against another (or more than one) bitbucket ... it should be good enough to tweak the provision app's CM to point to the new bitbucket / jira ... host.

This will ensure that the resources are created on the right bitbucket (e.g. project / repo) - BUT the jenkins job / webhook proxy that is called, does not get the bitbucket URL as param, but rather takes the one configured within jenkins.. (which is the "default" cluster configured one)

https://github.com/opendevstack/ods-core/blob/master/create-projects/Jenkinsfile#L16

based on
 
https://github.com/opendevstack/ods-provisioning-app/blob/master/src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java#L374-L389

@michaelsauter  / @stitakis  - did I see this correct?

@metmajer fyi

stitakis:
partially fixes #407

### DIFF
--- a/src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
+++ b/src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
@@ -61,6 +61,8 @@ public class JenkinsPipelineAdapter extends BaseServiceAdapter implements IJobEx
 
   public static final String PROJECT_ID_KEY = "PROJECT_ID";
 
+  public static final String OPTION_KEY_GIT_SERVER_URL = "GIT_SERVER_URL";
+
   @Value("${openshift.jenkins.webhookproxy.name.pattern}")
   protected String projectOpenshiftJenkinsWebhookProxyNamePattern;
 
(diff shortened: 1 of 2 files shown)


stitakis:
I still need to do more testing

clemensutschig:
@stitakis  - super nice! .. I would assume that the jenkins is then also configured against that bitbucket 
1. shared libraries
1. bitbucket host env params

Can you confirm?

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
```diff
@@ -61,6 +61,8 @@
 
   public static final String PROJECT_ID_KEY = "PROJECT_ID";
 
+  public static final String OPTION_KEY_GIT_SERVER_URL = "GIT_SERVER_URL";
```
clemensutschig, line 64: @michaelsauter  is there a hidden knob in the webhook proxy host, to cater to this name?

stitakis:
@clemensutschig @michaelsauter please have a look at the 2nd commit

clemensutschig:
@stitakis  -> master?

PR #426, opened 2020-04-28
update ocp-config after adding prov-app to ods-core

oalyman:


openshift/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+To apply the OCP config for `prov-dev`:
+```
+tailor update -n prov-dev
+```
+
+To apply the OCP config for `prov-test`:
+```
+tailor update -n prov-test
+```
```
michaelsauter, line 14: It would be nice to explain the configuration procedure quickly.

openshift/Tailorfile
```diff
@@ -1,7 +1,6 @@
 namespace prov-dev
 selector app=prov-app
-param-file ../../../ods-configuration/ods-core.env
-param-file ../../../ods-configuration/ods-provisioning-app-dev.env
+param-file ods-provisioning-app.env
```
michaelsauter, line 3: Should that file be gitignored?

clemensutschig:
@oalyman.. This needs adapts.. especially with the new configurable ods namespace..

michaelsauter:
@oalyman Can you also add in the changes similar to https://github.com/opendevstack/ods-core/pull/535? I don't want to do them in a PR here now because that would only create more merge conflicts ...

openshift/Tailorfile
oalyman: No need to add it as *.env files are already ignored.

https://github.com/opendevstack/ods-provisioning-app/blob/3d9370053b54fee2961bd90cf47a5ec53e7bd385/.gitignore#L33

openshift/README.md
oalyman: done

PR #432, opened 2020-05-05
Change readable repos config

### LINKED ISSUE #412: Improve configuration of readable repositories
If a project specific CD user is used, then it is given access to a list of repositories under the `opendevstack` Bitbucket project. That list is configured in https://github.com/opendevstack/ods-provisioning-app/blob/master/ocp-config/prov-app/cm.yml#L131-L133.

There are a couple of issues with this:

1. By default, it does not contain the `ods-document-templates` needed for the MRO when documents should be generated.
2. The list is hardcoded, making it hard to use `tailor diff/apply` against it when someone changes the list in the web console. Unfortunately, it is hard to change it to be parameterised right now, because OpenShift Templates (which is what Tailor uses under the hood) do not allow to parameterise multiple lines easily.

I think the most pragmatic solution would be to change:
```
scm.global.readablerepos.opendevstack[0]=ods-jenkins-shared-library
scm.global.readablerepos.opendevstack[1]=ods-quickstarters
scm.global.readablerepos.opendevstack[2]=ods-mro-jenkins-shared-library
```
to:
```
scm.global.readablerepos.opendevstack=${READABLE_OPENDEVSTACK_REPOS}
```
where `READABLE_OPENDEVSTACK_REPOS=ods-jenkins-shared-library,ods-quickstarters,ods-mro-jenkins-shared-library`. Then the prov app would need to transform into a `List`.

Thoughts or alternative suggestions @oalyman @stitakis?

michaelsauter:
That way it is easier to manage via OpenShift templates.

Closes #412.

### DIFF
--- a/ocp-config/prov-app/cm.yml
+++ b/ocp-config/prov-app/cm.yml
@@ -128,9 +128,7 @@ objects:
       bitbucket.technical.user=${CD_USER_ID}
       bitbucket.opendevstack.project=opendevstack
 
-      scm.global.readablerepos.opendevstack[0]=ods-jenkins-shared-library
-      scm.global.readablerepos.opendevstack[1]=ods-quickstarters
-      scm.global.readablerepos.opendevstack[2]=ods-mro-jenkins-shared-library
+      scm.global.readablerepos.opendevstack=ods-jenkins-shared-library,ods-quickstarters
 
       # ODS properties
       ods.image-tag=${ODS_IMAGE_TAG}
--- a/README.md
+++ b/README.md
@@ -3,24 +3,20 @@
 ![](https://github.com/opendevstack/ods-provisioning-app/workflows/Provisioning%20App%20Build/badge.svg?branch=master)
 
 ## Introduction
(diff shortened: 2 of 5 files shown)


ocp-config/prov-app/cm.yml
```diff
@@ -128,9 +128,7 @@ objects:
       bitbucket.technical.user=${CD_USER_ID}
       bitbucket.opendevstack.project=opendevstack
 
-      scm.global.readablerepos.opendevstack[0]=ods-jenkins-shared-library
-      scm.global.readablerepos.opendevstack[1]=ods-quickstarters
-      scm.global.readablerepos.opendevstack[2]=ods-mro-jenkins-shared-library
+      scm.global.readablerepos.opendevstack=ods-jenkins-shared-library,ods-quickstarters
```
michaelsauter, line 131: Removing `ods-mro-jenkins-shared-library` as it is no longer needed since it is merged into `ods-jenkins-shared-library`.

PR #438, opened 2020-05-11
add separate configuration for admin webhookproxy

oalyman:
Adds a separate configuration for the admin webhookproxy that proxies the create and delete project jobs defaulting to `webhook-proxy-ods`.

For the project webhook proxies that are created with each new project the pattern is still the same `webhook-proxy-${PROJECT_ID}-cd`

### DIFF
--- a/src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
+++ b/src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
@@ -65,8 +65,11 @@ public class JenkinsPipelineAdapter extends BaseServiceAdapter implements IJobEx
 
   public static final String OPTION_KEY_GIT_SERVER_URL = "GIT_SERVER_URL";
 
-  @Value("${openshift.jenkins.webhookproxy.name.pattern}")
-  protected String projectOpenshiftJenkinsWebhookProxyNamePattern;
+  @Value("${openshift.jenkins.admin.webhookproxy.host}")
+  protected String adminWebhookProxyHost;
+
+  @Value("${openshift.jenkins.project.webhookproxy.host.pattern}")
+  protected String projectWebhookProxyHostPattern;
 
   @Value("${openshift.jenkins.trigger.secret}")
   private String projectOpenshiftJenkinsTriggerSecret;
@@ -382,11 +385,7 @@ public class JenkinsPipelineAdapter extends BaseServiceAdapter implements IJobEx
     if (jenkinsPipelineProperties.isAdminjob(job.getId())) {
 
       boolean deleteComponentJob = jenkinsPipelineProperties.isDeleteComponentJob(job.getId());
-      String webhookProxyHost =
-          String.format(
-              projectOpenshiftJenkinsWebhookProxyNamePattern,
-              deleteComponentJob ? projID : "prov",
-              projectOpenshiftBaseDomain);
+      String webhookProxyHost = computeWebhookProxyHost(job.getId(), projID);
       String url =
           buildExecutionUrlAdminJob(
               job, componentId, projID, webhookProxySecret, webhookProxyHost, deleteComponentJob);
@@ -396,9 +395,7 @@ public class JenkinsPipelineAdapter extends BaseServiceAdapter implements IJobEx
       execution.project = bitbucketOdsProject;
 
     } else {
-      String webhookProxyHost =
-          String.format(
-              projectOpenshiftJenkinsWebhookProxyNamePattern, projID, projectOpenshiftBaseDomain);
+      String webhookProxyHost = computeWebhookProxyHost(job.getId(), projID);
 
       execution.url =
           JenkinsPipelineAdapter.buildExecutionUrlQuickstarterJob(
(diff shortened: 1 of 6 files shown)


src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
```diff
@@ -65,8 +65,11 @@
 
   public static final String OPTION_KEY_GIT_SERVER_URL = "GIT_SERVER_URL";
 
-  @Value("${openshift.jenkins.webhookproxy.name.pattern}")
-  protected String projectOpenshiftJenkinsWebhookProxyNamePattern;
+  @Value("${openshift.jenkins.admin.webhookproxy.host}")
+  protected String adminWebhookProxyHost;
```
stitakis, line 69: any reason why not use `private` instead?

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
```diff
@@ -65,8 +65,11 @@
...
+
+  @Value("${openshift.jenkins.project.webhookproxy.host.pattern}")
+  protected String projectWebhookProxyHostPattern;
```
stitakis, line 72: any reason why not use `private` instead?

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
```diff
@@ -416,6 +413,16 @@ private Execution buildExecutionObject(
     return execution;
   }
 
+  private String computeWebhookProxyHost(String jobId, String projID) {
```
stitakis, line 416: are these 3 cases in this method unit tested?

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
oalyman: It's aligned with the visibility of the other attributes in this class.

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
oalyman: It's aligned with the visibility of the other attributes in this class.

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
oalyman: As the change only modifies the existing functionality the existing integration / e2e tests were adapted and test the functionality.
See the changes in:
```
org.opendevstack.provision.controller.E2EProjectAPIControllerTest#testProvisionNewSimpleProjectInternal
org.opendevstack.provision.controller.E2EProjectAPIControllerTest#testQuickstarterProvisionOnNewOpenProjectInclDeleteWholeProject
org.opendevstack.provision.controller.E2EProjectAPIControllerTest#testQuickstarterProvisionOnNewOpenProjectInclDeleteSingleComponent
```

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
stitakis: ok, then can you please change it to `private`?

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
stitakis: ok, then can you please change it to `private`?

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
oalyman: It's not possible bcs. of how the Test is written. See lines starting from https://github.com/opendevstack/ods-provisioning-app/blob/7db613174625a454dc1a7ed9b9ca605b5886e9a5/src/test/java/org/opendevstack/provision/services/JenkinsPipelineAdapterTest.java#L73

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
oalyman: It's not possible bcs. of how the Test is written. See lines starting from https://github.com/opendevstack/ods-provisioning-app/blob/7db613174625a454dc1a7ed9b9ca605b5886e9a5/src/test/java/org/opendevstack/provision/services/JenkinsPipelineAdapterTest.java#L73

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
stitakis: ok, actually this is an anitpattern in java... the preferred way is to use getters and setters in such cases. We can leave it this way for now.

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
stitakis: ok, actually this is an anitpattern in java... the preferred way is to use getters and setters in such cases. We can leave it this way for now.

PR #447, opened 2020-05-15
enables docker cmd to instruct spring to load properties from additional location

### LINKED ISSUE #446: reuse application properties from src/main/resources in openshift to simplify configuration
Simplify configuration by reusing application properties.
Currently application properties contained in src/main/resources are fully duplicated in cm.yml.
This is because the docker file override the `spring.config.location` property to just one location `--spring.config.location=classpath:/,file:/config/`.
This is wrong. Composite configuration based on default provided by application.properties and overrides provided in file:/config/ should be allowed to facilitate property overrides over time.  
Currently only 1/3 of the configuration is overriden and parameterized.

Change in Dockerfile, the spring config location from:
`CMD ["java", "-jar", "app.jar", "--spring.config.location=file:/config/"]`
to
`CMD ["java", "-jar", "app.jar", "--spring.config.location=classpath:/,file:/config/"]`
Adding `classpath:/` before `file:/config/` will instruct spring to load the properties first from the first location `classpath:/` and the overrided from the 2nd location `file:/config/`.

The `cm.yml` will need to be modified to include only the application configuration overrides.

stitakis:
fixes #446
This change leave the configuration overrides in the cm.yml
need to be followed up by a change in ods-core `ods-provisioning-app/ocp-config/cm.yml`

### DIFF
--- a/openshift/cm.yml
+++ b/openshift/cm.yml
@@ -67,6 +67,12 @@ parameters:
 - name: READABLE_OPENDEVSTACK_REPOS
   required: true
   description: Repositories which the project specific technical user can read
+- name: WEBHOOK_PROXY_EVENTS
+  required: true
+  description: list of supported webhook events
+- name: PROV_APP_WEBHOOKPROXY_HOST
+  required: true
+  description: the webhookproxy host that prov app connect to
 objects:
 - kind: ConfigMap
   metadata:
@@ -77,10 +83,6 @@ objects:
   data:
     properties: |
       # log level
-      logging.level.root=WARN
-      logging.level.org.springframework=INFO
-      logging.level.org.springframework.security=DEBUG
-      logging.level.com.atlassian=INFO
       logging.level.com.atlassian.crowd=${PROV_APP_LOG_LEVEL_ATLASSIAN_CROWD}
       logging.level.org.opendevstack=${PROV_APP_LOG_LEVEL_OPENDEVSTACK}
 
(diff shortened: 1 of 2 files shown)


openshift/cm.yml
```diff
@@ -77,219 +83,62 @@ objects:
...
-      confluence.json.rpc.api.path=/rpc/json-rpc/confluenceservice-v2
-      confluence.permission.filepattern=classpath*:permission-templates/confluence.permission.*
 
       # Bitbucket properties
       bitbucket.uri=${BITBUCKET_URL}
-      bitbucket.api.path=/rest/api/1.0
-      bitbucket.repository.pattern=%s-%s
       bitbucket.default.user.group=${PROV_APP_CROWD_USER_GROUP}
       bitbucket.technical.user=${CD_USER_ID}
-      bitbucket.opendevstack.project=opendevstack
```
michaelsauter, line 132: Could you not remove this, but instead parameterise it by setting it to https://github.com/opendevstack/ods-core/blob/master/configuration-sample/ods-core.env.sample#L19? Thanks!

openshift/cm.yml
```diff
@@ -67,6 +67,12 @@ parameters:
 - name: READABLE_OPENDEVSTACK_REPOS
   required: true
   description: Repositories which the project specific technical user can read
+- name: WEBHOOK_PROXY_EVENTS
+  required: true
+  description: list of supported webhook events
+- name: PROV_APP_WEBHOOKPROXY_HOST
```
michaelsauter, line 73: I don't find this parameter in https://github.com/opendevstack/ods-core/blob/master/configuration-sample/ods-core.env.sample?

oalyman:
@stitakis What is the reason for this change? With the previous config it already loaded all properties in the /config folder.

oalyman:
@stitakis adding `classpath:/` is IMO dangerous as the distributed jar contains *.properties files.
and additionally I don't understand why you want to load properties from the classpath at all.

oalyman:
@stitakis Ok found that you want to distribute the existing configurations as default (#446)

It bears the risk that you might have unintended configuration (defaults) in the target environment when you miss to define them for the target environment.
I would always prefer a setup where starting the application fails if some configuration is not defined.

stitakis:
canceled this in favour of #473

Issue #363, opened 2020-02-27
Separate quickstarter config from application.properties


michaelsauter:
**Is your feature request related to a problem? Please describe.**
At the moment, the `ConfigMap application.properties` defines which quickstarters are available: https://github.com/opendevstack/ods-provisioning-app/blob/master/ocp-config/prov-app/cm.yml. This is not ideal as it does not allow to compare the content of the file easily via Tailor when users change which quickstarters they want to use (e.g. if they point to internal quickstarters).

**Describe the solution you'd like**
It would be good to have the quickstarter configuration separately. This could be another ConfigMap, or a file. A ConfigMap has the advantage to be easily modifiable in the OCP UI.

stitakis:
@michaelsauter  I first solution that comes to my mind is to move the current jenkinspipeline/quickstarters definition to a file (defined in oc as configmap) and modify the ProvApp to load this configuration from a default location/file. The file format could be json for easy convertion to the current `Quickstarter` java class/model.
Would this address the requirements, especially related to Tailor?

michaelsauter:
Some questions:
* Where would this file be stored?
* How could we ship with a default set?

Since the file should be edited by humans, I suggest YAML instead of JSON.

One alternative I have thought about (though not sure yet if this is a good idea): One could have a ConfigMap, which is mounted as a file. Tailor has a flag called `--preserve`, which allows to preserve paths of live configuration when applying. One could use that to create the ConfigMap initially with some default quickstarters, but subsequent calls of `tailor apply` would not show drift in the ConfigMap.

michaelsauter:
Here's my proposal for this, which I believe is the least effort to get it done for v3:

Add a `ConfigMap` like this:
```
apiVersion: v1
kind: Template
objects:
- apiVersion: v1
  kind: ConfigMap
  metadata:
    labels:
      app: ods-provisioning-app
    name: quickstarters
  data:
    properties: |
      be-java-springboot:
        desc: Java SpringBoot
        repo: ods-quickstarters
      be-golang-plain:
        desc: Go
        repo: ods-quickstarters
```

Then, mount this file in the prov app, and read the YML. The structure of the YML is just an example, we might also use a top-level key ... not sure there.

Further, we ensure that this `ConfigMap` does not detect drift on its data. That allows to seed with defaults, but when users adapt the config, no drift is shown.

To achieve this with Tailor, we add the following line to the `Tailorfile`:
```
preserve cm: quickstarters:/data
```

[...]

PR #448, opened 2020-05-19
add gateway kong quickstarter configuration

stefanlack:
Adds configuration for new quickstarter. See https://github.com/opendevstack/ods-quickstarters/pull/291

### DIFF
--- a/src/main/resources/application.properties
+++ b/src/main/resources/application.properties
@@ -147,6 +147,9 @@ jenkinspipeline.adminjobs.delete-components.repo=ods-core
 jenkinspipeline.quickstarter.be-java-springboot.desc=Backend - SpringBoot/Java
 jenkinspipeline.quickstarter.be-java-springboot.repo=ods-quickstarters
 
+jenkinspipeline.quickstarter.be-gateway-kong.desc=Backend - Gateway Kong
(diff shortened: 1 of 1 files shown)


stitakis:
Hi @stefanlack we are in the process of separating the quickstarters configuration from the application.properites in issue #363. In this context, do you think we need to add kong to the default list of quickstarters or would you prefer to keep it as part of your own deployment configuration on your side? @michaelsauter, your thoughts?

stefanlack:
Since we are still on the 2.x version, the change in issue #363 has no effect for us. Since Kong Quickstart is still not migrated to 3.x version, feel free to not add the quickstarter to the default list.

If the migration is done and the kong quickstarter is available for 3.x version, we can add it later to the default configuration.

stitakis:
@stefanlack who will be the active mantainer of kong qs? Asking because we are aiming to add to the default configuration and keep in `ods-quickstarters` only the quickstarters with active maintainers

stefanlack:
As written in https://github.com/opendevstack/ods-quickstarters/pull/291#issuecomment-635272931, the opendevstack team at opitz-consulting is not going to migrate to 3. in the near future. Therefore, we are not able to maintain the quickstarter, I am sorry to say.

PR #449, opened 2020-05-25
splits qs configuration from application.configuration

### LINKED ISSUE #363: Separate quickstarter config from application.properties

stitakis:
fixes #363


openshift/cm.yml
```diff
@@ -293,3 +245,67 @@ objects:
       spring.mail.password=${PROV_APP_MAIL_PASSWORD}
       provison.mail.sender=provision@${PROV_APP_MAIL_HOST}
       spring.main.allow-bean-definition-overriding=true
+
+- kind: ConfigMap
+  metadata:
+    labels:
+      app: prov-app
+    name: quickstarters.properties
```
michaelsauter, line 253: You'd then need to add `preserve bc:quickstarters.properties:/data` to the `Tailorfile` so no drift is generated for this.

oalyman:
@michaelsauter @stitakis LGTM

michaelsauter:
@stitakis Should we move forward with this? The feature is quite crucial and I don't see anything that would block your change?

stitakis:
@michaelsauter we can move forward, do you approve this change?

stitakis:
closing this in favour of #458

PR #457, opened 2020-06-02
adds option to disable confluence adapter by config

### LINKED ISSUE #450: disable confluence adapter per configuration
Problem:
Not all installation of ProvApp requires confluence to be available.

But currently there is no configuration option that enables to deploy ProvApp with a disabled confluence adapter.

Proposed solution:
Add a spring property that makes possible to disable the confluence adapter, so that the `ProjectApiController.java` do not use the `confluenceAdapter` at runtime.

stitakis:
fixes #450


stitakis:
@metmajer @georgfedermann  this makes posible to provision project without confluence... useful for ODS DevEnv if confluence is not available there

PR #458, opened 2020-06-02
splits qs configuration from application.configuration

### LINKED ISSUE #363: Separate quickstarter config from application.properties

stitakis:
fixes #363


michaelsauter:
Hmm, GitHub shows me a few unrelated commits and a merge conflict. To me, it looks like you'd want a branch that only contains fcc1c9e6d86714fc0b3746f8ad0fc9b487bf43a6, and then one would only need to fix the merge conflicts. Could you adapt please?

PR #469, opened 2020-06-15
get back be-gateway-nginx quickstarter

gerardcl:
recovering removed be-gateway-nginx quickstarter from cm.yml

### DIFF
--- a/ocp-config/cm.yml
+++ b/ocp-config/cm.yml
@@ -265,6 +265,9 @@ objects:
   data:
     properties: |
       # Backend Quickstarter
+      jenkinspipeline.quickstarter.be-gateway-nginx.desc=Backend - Gateway/Nginx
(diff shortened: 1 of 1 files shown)


michaelsauter:
Actually, you should add the quickstarter here: https://github.com/opendevstack/ods-core/blob/master/ods-provisioning-app/ocp-config/cm.yml. Otherwise the gateway quickstarter will be missing in a fresh ODS install.

I'm wondering if it would be better to just make one example in the file you changed here. Otherwise we have to maintain two lists and that will never work out really well? :)

E.g. we could just list the "docker plain" quickstarter here, and then people add the ones they need for local development.

Thoughts @oalyman @gerardcl @stitakis?

gerardcl:
hi! I agree, since here the provisioning-app is ment for developing it one might not need to have by default all the quickstarters but, as you mention, just have an example config, hence it can then be changed/updated per need while developing it.

PR #473, opened 2020-06-22
fixes broken spring configuration: instruct docker to load properties…

### LINKED ISSUE #446: reuse application properties from src/main/resources in openshift to simplify configuration

stitakis:
… from additional location

fixes #446


src/main/resources/application-quickstarters.properties
```diff
@@ -0,0 +1,55 @@
+# Backend Quickstarter
```
michaelsauter, line 1: The name of this file seems to differ from what you load above?

stitakis: yes, this is on purpose, `src/main/resources/application-quickstarters.properties` defines a spring profile. It can can be optionally enabled i.e. when running the application locally.

stitakis:
@michaelsauter thanks a lot :+1:!

clemensutschig:
My hunch is that a file:/ is missing.. @stitakis.. we need a fix for this asap. And a solid Test..

stitakis:
@clemensutschig I tested it locally and in octest cluster. I'm taking a look at the problem in ODSBox reported by #477

PR #482, opened 2020-06-26
adds error codes to create project precondition checks

### LINKED ISSUE #479: Add error codes to the preconditions check response
Currently the payload returned by the preconditions checks do not include an error code.
For easier integration with other system we need to add error codes to the json response.
This will allow other systems calling the `add project` endpoint to add retry logic for specific error codes.
This is a follow-up of #330

stitakis:
fixes #479


clemensutschig:
@stitakis .. nice.. Please fix the checks that failed..

PR #497, opened 2020-07-02
enables spring security authorization for crowd profile

### LINKED ISSUE #172: Protect API operations better
today the API operations are just authentication protected thru auth interceptors - we should protect the api as well - in lieu of /delete coming

stitakis:
fixes #172


stitakis:
@renedupont thanks a lot for the review... more PR are coming new week! ;-)

PR #502, opened 2020-07-06
Pass ODS namespace to Jenkins pipeline

michaelsauter:


src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
```diff
@@ -111,11 +111,14 @@
   @Value("${bitbucket.opendevstack.project}")
   protected String bitbucketOdsProject;
 
+  @Value("${ods.namespace}")
+  protected String odsNamespace;
+
   @Value("${ods.image-tag}")
-  private String odsImageTag;
+  protected String odsImageTag;
```
stitakis, line 118: In java the most common practice is to use getter and setter methods for accessing internals/class members. This is because many reason, related to encapsulation but also byte decorators and mocking libraries that create proxies in behind.
Can please add getter and setters for `odsGitRef` and `odsImageTag` and access to those class members throu them?

michaelsauter: I just followed how the other properties are tested in this class - in order to have a test for the `ods` properties at all. I agree that this is not the best way to do things, but refactoring is a separate topic - please open a new issue. Can we proceed with merging?

stitakis: @michaelsauter from my point of view is not a refactoring topic... this change changes the visibility of the current properties from `private` to `protected` and break the encapsulation. The test functionality can be achived with the getter and setter method. 
If you take a look at the commit story, you will see that I'm fixing the wrongly defined as `protected` properties.

michaelsauter: Merged - follow-up issue at #503.

PR #506, opened 2020-07-08
Feature/gh 504 enable basic auth by default

### LINKED ISSUE #504: enable basic auth beside crowd authentication for easy endpoint consumption
When crowd authentication is enabled REST API calls need to include cookies like session id and crowd session key. 
To get this cookies an "authentication" request needs to be done as first request. 

This work fine for the front-end communication.
But for the REST API endpoint consumption in integration and test automation scenarios adds a bit of complexity.

Basic Auth will provide an easier way to authenticate request.

stitakis:
fixes #504


stitakis:
@clemensutschig fyi... this fixes the problem with basic auth failing at first login

PR #517, opened 2020-07-14
Feature/gh x add spa ui

stitakis:
fixes #517


stitakis:
cancelling in favour of #519

Issue #518, opened 2020-07-14
Provide modern SPA user interface to improve UX


stitakis:

netzartist:
The current UI lacks a good usability which leads to misuse especially for non-technical users and requires extra caretaking when entering/editing data.

**Main goals:**

- Improve the overall user experience incorporating latest usability best practices
- Provide a modern tech stack to allow long-term maintainability and keep high code-quality 

**The new UI layout** consists of a:

- Sidebar to quickly navigating to an existing project or create a new project
- Detail view showing all information compact and clear including component's data (via expand/collapse toggle) and a switch between edit-/view-mode

**Features:**

- The same like in the current UI, plus
- 'Delete project' functionality
- 'Delete component' functionality

![image](https://user-images.githubusercontent.com/370980/87682141-c9490000-c77f-11ea-8c79-848e13611fcd.png)

As there are some open tasks to be finished, the new UI will be included as an experimental feature in ODS 3.x release, deactivated by default. It can be activated by an admin with a `frontend.spa.enabled` feature flag in Spring Boot.

PR #519, opened 2020-07-14
Feature/copy nfe be static resources

stitakis:


clemensutschig:
@stitakis  - we need the jenkinsfile additions here as well - or how else would be build work?

stitakis:
@clemensutschig yes, I know... plan is to add a working Jenkinsfile (that compile the new FE in a different jenkins node) in a separate change. 

Currently the ProvApp is built in github (github workflow) and publish to dockerhub. 

This PR could be merged once we figure out the changes related to compile the FE in github workflow. I'm trying to enable this... but my last commit in this branch didn't worked.

stitakis:
The FE build is working in github, this PR is ready for review!

stitakis:
@michaelsauter @cschweikert @clemensutschig any comments... do you approve this change?

michaelsauter:
@stitakis My understanding is:

* The FE assets are built on GitHub and stored inside the image published on DockerHub?
* The new FE is not visible yet (question: how to enable it?)
* The old FE is not touched at all

Based on this I am fine with the general change (I can't say too much regarding the FE code ...)

I would appreciate if we had some screenshots in this PR or the issue to give an idea of how it looks. Alongside an explanation how to enabled / disable. E.g. is this something an admin does for all - or can the users switch on a case-by-case basis?

netzartist:
@michaelsauter Will add the context, description and screenshots to #518. Each commit of this PR has run through PR approvals already. Might squashing all into 1 commit **before** the merge help to a better glimpse on the (mostly) FE code?

stitakis:
@michaelsauter thanks for taking a look at this!

Answers to questions

> `The FE assets are built on GitHub and stored inside the image published on DockerHub?`

Yes, the `gradle.build` was modified to copy the new FE assets into the `app.jar` file

> `The new FE is not visible yet (question: how to enable it?)`

Exactly, the BE will serve the new FE if the flag `frontend.spa.enabled` set to `true` (default value is `false`)
@netzartist can you please add this to the documentation?

> `The old FE is not touched at all`

Exactly, no changes in old FE

netzartist:
Added description and screenshot here: https://github.com/opendevstack/ods-provisioning-app/issues/518#issuecomment-659464786

clemensutschig:
I have "bauchweh" with this ... this will break the idea of cloning locally and working on it based on ODS magic .. what holds us back on making the build(s) work via ODS? ... just asking...

stitakis:
@clemensutschig this PR ships a `Jenkinsfile` compatible with ODS 3.x. It has only one missing part that needs to be added, which is the compilation of the new FE (I propose to add this capability in a different change).

I'm not sure if I'm understand your last comment... since we build the ProvApp in github and push it to dockerhub, the corresponding build logic is defined in `.github/workflow` and `build.gradle`. However the ProvApp still includes support for ODS/Jenkins build defined in `Jenkinsfile`.

Does this answer your question?

michaelsauter:
@clemensutschig We will look into adapting the `Jenkinsfile` so that one can still build an image from scratch.

Issue #516, opened 2020-07-14
Add project specific cd user after for generating new components


sino92:
Creation of new project with a project-specific-cd-user is a bit clumsy as prov-app doesn't communicate with Crowd or AD to create one in real time. For AD, I assume there can't be instant feedback and response to create a new technical user anyway.
As a result of this lag, there can be a use case where project specific cd user is created after the project is already generated. 

It would save developers time if this new user can then be added when generating new components. This would link the generated repo to that cd user as opposed to global cd user.
@gerardcl @stitakis

michaelsauter:
Right now there can only be one "cd_user" per project (foo-cd namespace). Do you suggest to change that when generating a new component, or is your suggestion to have multiple "cd_user"s per project?

sino92:
Sorry, a correction for above... I meant to say switching from global cd user to a project specific one, not adding a new one. So saving the developer time to change related resources:
- changing OC username/secret
- changing the user in project repo  
- adding a new user to opendevstack quickstarters and jenkins shared lib repos

michaelsauter:
FYI

> changing OC username/secret

that was - at the time - a conscious design decision. We intentionally kept the password out of ODS altogether, and we also avoided integrating ODS with the "user management process" (as that can differ a lot between companies). So unless we give that up, changing the secret manually is painful, but the most secure option right now.

sino92:
Hm makes sense as a design decision. Then maybe we can add some documentation around which places need change in case someone need to change their cd user in the middle of the project. What do you think?

michaelsauter:
Yes that would be great!

PR #570, opened 2020-08-11
Add info about cd user

sino92:
https://github.com/opendevstack/ods-provisioning-app/issues/516


docs/modules/provisioning-app/pages/cd-user.adoc
```diff
@@ -0,0 +1,17 @@
+:toc: macro
+
+= Syncing your Technical User With ODS Components
+
+When you create a new project, the provisioning app triggers a Jenkins pipeline. This pipeline adds `CD_USER_ID` as the name and `CD_USER_B64` as the encoded password to the secrets of your newly generated Openshift cd project. This is the technical user for your project. Every time a (cd) project is generated, these parameters are added to the secrets of your Openshift project. At this point however, you're still not ready to push some code into your repository or succesfully generate a new quickstarter using ods-shared-library. In order to achieve the two, you would manually need to add this same technical user to the repository settings of the two ODS repositories: `ods-jenkins-shared-library` and `ods-quickstarters`. Hence you have a technical user that will have read access to your project.
```
michaelsauter, line 5: Typo `CD_USER_B64`

michaelsauter: > In order to achieve the two, you would manually need to add this same technical user to the repository settings of the two ODS repositories: `ods-jenkins-shared-library` and `ods-quickstarters`.

The prov app gives read access to the project specific cd user. See `READABLE_OPENDEVSTACK_REPOS`. Are you talking about the global cd user? I would discourage to use that.

docs/modules/provisioning-app/pages/cd-user.adoc
```diff
@@ -0,0 +1,17 @@
+:toc: macro
+
+= Syncing your Technical User With ODS Components
+
+When you create a new project, the provisioning app triggers a Jenkins pipeline. This pipeline adds `CD_USER_ID` as the name and `CD_USER_B64` as the encoded password to the secrets of your newly generated Openshift cd project. This is the technical user for your project. Every time a (cd) project is generated, these parameters are added to the secrets of your Openshift project. At this point however, you're still not ready to push some code into your repository or succesfully generate a new quickstarter using ods-shared-library. In order to achieve the two, you would manually need to add this same technical user to the repository settings of the two ODS repositories: `ods-jenkins-shared-library` and `ods-quickstarters`. Hence you have a technical user that will have read access to your project. 
+
+== Setting Your Project Specific CD User
+
+Keep in mind that at this point, we assume you've already created at least two ODS projects that use the same technical user and you'd like to change one of them. When you make this change, you need to have this new technical user be in sync in three places:
```
michaelsauter, line 9: Why two? One is enough?

docs/modules/provisioning-app/pages/index.adoc
```diff
@@ -14,6 +14,7 @@ It delegates the tasks to create / update resources to several services such as
...
+ ** If you've set up your ODS infrastructure properly, then you've probably already filled up the cd user fields here in our ods-core environment file link:https://github.com/opendevstack/ods-core/blob/master/configuration-sample/ods-core.env.sample#L137[sample]. Once you create a new project, it adds this cd user credential information you provided to the cd component of your project. This is the technical user that will communicate with the ODS components of your fork. In order for your project to work in sync with ODS components, it needs to have read permissions over two ODS repositories: `ods-jenkins-shared-library` and `ods-quickstarters`. The read permission for shared library is needed so that the library can checkout your changes and deploy them into designated Openshift environment. As for the ladder ODS repository, when you generate a new component it needs to access the chosen quickstarter repository and execute it's Jenkinsfile. If you don't want a single cd user to have read access to all your projects, you can add a new project specific cd user that works in sync with ODS components. More information can be found xref:cd-user.adoc[here].
```
michaelsauter, line 17: > Once you create a new project, it adds this cd user credential information you provided to the cd component of your project

Only true if "project specific user" is not selected - which it should normally be.

docs/modules/provisioning-app/pages/cd-user.adoc
sino92: The reason why I said two was because the second project using the same credentials as the first one being a deterrent to the developer. It wasn't clear from here that I was encouraging the developer to use separate cd user credentials for 1+...

docs/modules/provisioning-app/pages/index.adoc
sino92: Why not make project specific the default then?

docs/modules/provisioning-app/pages/cd-user.adoc
sino92: You're right. I meant to say `Hence you have a technical user that ODS components can refer to in order to access to your project`, as the project read permission alone isn't enough. Also could you point out where and how `READABLE_OPENDEVSTACK_REPOS` is used in prov-app?

docs/modules/provisioning-app/pages/cd-user.adoc
sino92: It makes sense to strongly encourage the developer not to use global cd user as you pointed out

PR #578, opened 2020-08-19
adds snyk scanning github actions

stitakis:


clemensutschig (changes requested):
@stitakis - Please also upload the scan report similar to other Actions.

stitakis:
@michaelsauter not sure why is not working... I'll start to suspect that the imported snyk action does not work properly... just to double, can you please have a quick look if the way I access the secrets is correct?

michaelsauter:
@stitakis After looking at it again, I just realised - you're PR is from a branch in BIX-Digital. Secrets are not shared with forks ... push your branch to the OpenDevStack org, and it should work.

clemensutschig (approved):

stitakis:
@michaelsauter I pushed this branch to OpenDevStack.org and worked. Thanks for the hint.
If I add the same secret to the fork BIX-Digital, the secret should be available to this branch, right?

michaelsauter:
@stitakis Give it a try :) Would be nice to have.

stitakis:
> @stitakis - Please also upload the scan report similar to other Actions.

@clemensutschig I can't figure out which similar action do you mean... can you please point me to one?

Issue #527, opened 2020-07-20
Refactor error handling


netzartist:
- Centralize error types
- Map http code numbers to enums
- Manage app-specific error codes

Related:
- https://medium.com/angular-in-depth/expecting-the-unexpected-best-practices-for-error-handling-in-angular-21c3662ef9e4
- https://github.com/prettymuchbryce/http-status-codes/blob/master/index.d.ts
- https://blog.angular-university.io/rxjs-error-handling/
- https://dev.to/buildmotion/angular-errorhandler-to-handle-or-not-to-handle-1e7l

PR #634, opened 2020-11-18
Refactor SPA Error Handling

netzartist:
(Followup to #527)

Improved error handling and preparation for error logging to be added in a next step:

- `HttpRequestInterceptor` takes care of catching (and later logging) HTTP errors
- `GlobalErrorHandler` takes care of logging client errors

Logging should happen like this:

1. Send logs to a backend logging endpoint
2. Based on env settings passed to the Angular build script by an prov-app admin:
- Print client errors to the browser console and / or
- allow that the users see them "nice" in the app (e.g. via additional icon next to message which displays a modal dialog)

The current behaviour from a user's perspective is to see a meaningful message (without any tech info) while still being able to use the app. For example when for whatever reason the list of quickstarters cannot be loaded, then editing of a project is not possible and the user sees an information about this problem, but is still able to view all project details.

In a next step I'd prepare the possibility to configure an remote logging endpoint and create an Angular service which "fire & forgets" the logs to such an URL. 

What do you think? Or do you have other ideas? /cc @stitakis @michaelsauter


client/src/app/modules/error-handler/http-request-interceptor.ts
```diff
@@ -39,6 +34,16 @@ export class HttpRequestInterceptor implements HttpInterceptor {
 
     req = req.clone(httpOptions);
 
-    return next.handle(req);
+    return next.handle(req).pipe(
+      retry(2),
```
cschweikert, line 38: When I get it right, this is retrying any kind of failed request two more times automatically. This might also happen with partially successful requests and might lead to creation of even more broken parts or cause other issues on top. I would not do a retry automatically.

client/src/app/modules/sidebar/components/sidebar.component.html
```diff
@@ -4,6 +4,7 @@
 
 <ng-container *ngIf="isError" data-test-error>
   <div class="error">The project list could not be loaded currently</div>
+  <!-- TODO add a reload button -->
```
cschweikert, line 7: Usually this kind of error should never happen. If it happens frequently, something is broken and that something might also cause other thing to fail frequently. Since the whole application might be unstable, I would not integrate the error handling like this. I would rather show a global error, e.g. via popup or switching to an error page.

client/src/app/modules/error-handler/http-request-interceptor.ts
netzartist: You'e right, it"s better to leave it to the component to retry requests, so I removed it here. Will think about in which components this would make sense to tackle (incl. a `retryWhen` to not fire constantly).

client/src/app/modules/sidebar/components/sidebar.component.html
netzartist: In the previous implementation I thought it can happen that one endpoint works but the other not - guess this is not realistic, either the endpoints are all responding or using the app is pretty much senseless. So now the app component tries to get all projects and if this doesn't work a general error page will be shown.

cschweikert (approved):

PR #636, opened 2020-11-19
Add default project groups to project groups on project creation

### LINKED ISSUE #635: Permission default project groups on project creation in openshift
Currently the only way to permission groups on project creation is by setting special permission set.
The ProvApp passes the special permission groups to the webhook proxy/jenkins when it triggers the create project request.
This was only meant for project specific roles.

However, cluster administrator in corporate environment need a way to get access to the project in openshift.
For them to avoid doing this manually we need a way to automate this.
This feature request is about of adding a configuration option that will be used to setup default project groups.
With this functionality corporate cluster admins will be able to permission their admin group on project creation.

stitakis:
fixes #635

### DIFF
--- a/docs/modules/provisioning-app/pages/configuration.adoc
+++ b/docs/modules/provisioning-app/pages/configuration.adoc
@@ -57,6 +57,13 @@ The configuration for the permission sets are configured:
 . Bitbucket Project is provisioned with tight read & write roles
 . Openshift Project roles linked to the passed groups (`READONLY` - `view`, `ADMINGROUP` - `admin`, `USERS` - `edit`)
 
+Furthermore if you need to define default permission for openshift (e.g. to setup membership permission for cluster admins) you can add this to your application properties:
+```
+jenkinspipeline.create-project.default-project-groups=ADMINGROUP=<MY_CLUSTER_ADMIN_GROUP_NAME>
+```
+
+In case special permissions sets are defined this the default project groups will be appended to the lis of permissions sets.
+
 == Project/Space types based on templates
 
 The default jira / confluence project' types are defined in https://github.com/opendevstack/ods-provisioning-app/blob/master/src/main/resources/application.properties[src/main/resources/application.properties] - and correspondingly in the config maps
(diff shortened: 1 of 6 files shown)


stitakis:
@felipecruz91 this is the draft impl

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
```diff
@@ -127,7 +127,7 @@
   protected String generalCdUser;
 
   @Value("${jenkinspipeline.create-project.default-project-groups:''}")
-  private String clusterAdminsEntitlementGroups;
+  protected String clusterAdminsEntitlementGroups;
```
felipecruz91, line 130: @stitakis Needed to be `protected` to set this value from the unit test

stitakis: @felipecruz91 actually `private` is the correct one. For the unit test we just need to add the property to the `application-utest.properties`. This is the on used by the `JenkinsPipelineAdapter` class. (see line 56 -> `@ActiveProfiles("utest")` ). I'll do this

felipecruz91 (approved):

stitakis:
@michaelsauter to complete this change I'll create the corresponding one in ods-core that extends the ods configuration sample

stitakis:
@borja44 fyi

docs/modules/provisioning-app/pages/configuration.adoc
```diff
@@ -57,6 +57,13 @@ The configuration for the permission sets are configured:
 . Bitbucket Project is provisioned with tight read & write roles
 . Openshift Project roles linked to the passed groups (`READONLY` - `view`, `ADMINGROUP` - `admin`, `USERS` - `edit`)
 
+Furthermore if you need to define default permission for openshift (e.g. to setup membership permission for cluster admins) you can add this to your application properties:
+```
+jenkinspipeline.create-project.default-project-groups=ADMINGROUP=<MY_CLUSTER_ADMIN_GROUP_NAME>
```
michaelsauter, line 62: How would this be set in the `ConfigMap`? They are defined in https://github.com/opendevstack/ods-core/blob/3.x/ods-provisioning-app/ocp-config/cm.yml. I wonder where to set it so that no drift shows up during verification.

stitakis: @michaelsauter this is the PR for ods-core/3.x: https://github.com/opendevstack/ods-core/pull/889... what do you mean with "no drift shows up during verification"?

michaelsauter: Ah okay got it. Looks fine!

Issue #624, opened 2020-10-27
Show an error message when the component id does not match the regex expression


felipecruz91:
**Is your feature request related to a problem? Please describe.**
When creating a component through the Provisioning app, the component id must comply with [this regex expression](https://github.com/opendevstack/ods-provisioning-app/blob/3.x/src/main/java/org/opendevstack/provision/model/OpenProjectDataValidator.java#L36).

When a user introduces a component id, no message is shown suggesting what is wrong with the component id entered.

![image](https://user-images.githubusercontent.com/15997951/97333177-89fc7600-187b-11eb-96b9-a11ec0e804d3.png)

**Describe the solution you'd like**
It would be helpful for the user so see what is wrong in the component id so that it can fix it.

**Describe alternatives you've considered**
If the user enters a component id with digits, and the regex expression does not allow it, an error should appear below the component id indicating that digits are not allowed as part of the component id.

michaelsauter:
@netzartist I'd think the new frontend does this?

Maybe the action here is to switch to the new frontend and delete the old frontend?

stitakis:
@michaelsauter  @netzartist to delete the old frontend we need to discuss with @metmajer and his team first.

netzartist:
@michaelsauter Yes, the new frontend uses the same [regex](https://github.com/opendevstack/ods-provisioning-app/blob/13f626ce59f48bc090ef7c234322d249700225eb/client/src/app/modules/app-form/config/validation.json#L20) and shows a more detailed explanation: 

![image](https://user-images.githubusercontent.com/370980/97408731-cd4af900-18fc-11eb-9295-d81035695192.png)

@stitakis Let's make a plan to switch to the new frontend.

felipecruz91:
@stitakis @netzartist Is this going to be shipped as part of 3.1?

stitakis:
@felipecruz91 you introduced an input validation logic for this in the backend. That part is done. However we haven't done it yet in the current UI. Do you have time to take a look at it?

PR #641, opened 2020-11-25
Updated regex and error message for the component id

felipecruz91:
Refers to #624


client/src/app/modules/app-form/config/validation.json
```diff
@@ -17,10 +17,11 @@
   },
   "quickstarters": {
     "componentName": {
-      "regex": "(([A-Za-z][-A-Za-z]*)?[A-Za-z]){3,40}",
+      "regex": "^(?!tests)(([A-Za-z][-A-Za-z]*)?[A-Za-z]){3,40}",
       "errorMessages": [
         "Please enter 3-40 characters",
-        "Please use letters and dashes (optional)"
+        "Please use letters and dashes (optional)",
+        "Please avoid using reserved keywords e.g. tests"
```
netzartist, line 24: Either the message should be changed to `Please avoid using tests as a reserved keyword` or the regex should be really extended with a **list of keywords**, then the message makes sense again.

felipecruz91: It makes total sense - thanks for the feedback @netzartist

stitakis (dismissed):

stitakis (approved):

PR #644, opened 2020-11-26
Make cdUser field mandatory

### LINKED ISSUE #577: Remove global cd user and make project specific cd user default
Global cd user has two big downsides:
- security
- reliability: If one changes the password for the user then other projects that use it are affected by it
@michaelsauter @stitakis 

A reminder that with specific cd user, the user will have to modify Openshift project secret for cd user manually. A small effort for a bigger win.

netzartist:
Fixes #577 

Although the cd User doesn't seem to be regex-checked in the backend I thought it makes sense at least to do so in the frontend. I used this one: `([a-zA-Z][a-zA-Z_0-9]+)`

Based on that I think it makes sense to re-evaluate the regex rules, e.g. do we need more where currently are no checks and less or more strict ones.

### DIFF
--- a/client/src/app/domain/project.ts
+++ b/client/src/app/domain/project.ts
@@ -9,7 +9,7 @@ export interface ProjectData {
   projectKey: string;
   description?: string;
   webhookProxySecret?: string;
-  cdUser?: string;
+  cdUser: string;
   quickstarters?: ProjectQuickstarterComponentsData[];
   bugtrackerSpace: boolean;
   platformRuntime: boolean;
--- a/client/src/app/modules/app-form/config/validation.json
+++ b/client/src/app/modules/app-form/config/validation.json
@@ -2,17 +2,15 @@
   "project": {
     "name": {
       "regex": "([a-zA-Z][a-zA-Z_0-9]+)",
-      "errorMessages": [
-        "Please enter 2-80 characters",
-        "Please use numbers, letters and underscore"
-      ]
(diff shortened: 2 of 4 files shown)


cschweikert (approved):

michaelsauter:
I'm not sure about making it mandatory yet. I believe in #577 we only agreed on turning it on by default ...

stitakis:
Mandatory means for me that the corresponding backend API would not accept a request that do not include the cd user. I think we will not change this because this will have an impact in corporate environment. So I vote to go for turning it on by default in UI as Michael suggested.

netzartist:
turning on = making the field mandatory and always send it to the backend (no change there needed as long as it shouldn't be validated but just passed on in the project creation process)

michaelsauter:
@netzartist Is it still possible to submit the form without a specific cd user then?

netzartist:
@michaelsauter Request-wise yes ;) but not within the UI as the submit button is disabled.

michaelsauter:
@netzartist I would like to still be able to submit without it for now.

netzartist:
Ah, I thought it was about forcing the user to always enter a project specific user. If the field should stay optional then there's nothing to do in the frontend, so this PR can be closed. @sino92, confirm?

michaelsauter:
@netzartist I think the UI should default to a project specific user. The old UI did not do that. If the new UI already does this, then from my perspective, we should not do anything.

netzartist:
Closed, see https://github.com/opendevstack/ods-provisioning-app/issues/577#issuecomment-738017138

PR #651, opened 2020-12-11
Feature/update dependencies

renedupont:
- I added the nebula-lint gradle plugin to identify unused dependencies
- The change in RestClientTest.java was due to the fact that the latest okhttp version returns nullpointer exception instead of illegalargument, probably due to their migration to kotlin and thus changing the httpurl.parse method

Info: This branch was branched out from feature/update-spring-boot (#645)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -74,13 +74,13 @@ dependencies {
     testImplementation('org.springframework.restdocs:spring-restdocs-mockmvc')
     testImplementation('org.springframework.security:spring-security-test')
 
-    implementation("org.thymeleaf.extras:thymeleaf-extras-springsecurity5")
+    implementation('org.thymeleaf.extras:thymeleaf-extras-springsecurity5')
 
     // oauth2
-    implementation 'org.springframework.security:spring-security-oauth2-core'
-    implementation 'org.springframework.security:spring-security-oauth2-client'
-    implementation 'org.springframework.security:spring-security-oauth2-jose'
-    implementation 'org.springframework.security.oauth:spring-security-oauth2:2.5.0.RELEASE'
+    implementation('org.springframework.security:spring-security-oauth2-core')
+    implementation('org.springframework.security:spring-security-oauth2-client')
+    implementation('org.springframework.security:spring-security-oauth2-jose')
+    implementation('org.springframework.security.oauth:spring-security-oauth2:2.5.0.RELEASE')
 
     implementation(group: 'com.openshift', name: 'openshift-restclient-java', version: '9.0.1.Final') {
         exclude(group: 'org.slf4j', module: 'slf4j-api:')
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -100,16 +99,12 @@ dependencies {
...
-    implementation 'com.github.ulisesbocchio:jasypt-spring-boot-starter:2.1.1'
+    implementation 'com.github.ulisesbocchio:jasypt-spring-boot-starter:2.1.2'
 
     //easy http calls to atlassian JSON APIs
-    implementation "com.squareup.okhttp3:okhttp:4.4.0"
+    implementation "com.squareup.okhttp3:okhttp:4.9.0"
 
-    implementation group: 'commons-httpclient', name: 'commons-httpclient', version: '3.1'
-    implementation group: 'com.fasterxml.jackson.datatype', name: 'jackson-datatype-joda', version: '2.10.0.pr3'
-    implementation group: 'joda-time', name: 'joda-time', version: '2.10.4'
```
stitakis, line 110: @renedupont not sure if the removal of this dependencies is causing a `ClassNotFoundException`. 
I got this error in a development environment... we need to investigate it:
``` 
2020-12-16 10:43:54 ERROR o.a.c.c.C.[.[.[.[dispatcherServlet] - Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Filter execution threw an exception] with root cause
java.lang.ClassNotFoundException: org.apache.commons.httpclient.methods.RequestEntity
    at java.base/java.net.URLClassLoader.findClass(Unknown Source)
    at java.base/java.lang.ClassLoader.loadClass(Unknown Source)
    at org.springframework.boot.loader.LaunchedURLClassLoader.loadClass(LaunchedURLClassLoader.java:151)
    at java.base/java.lang.ClassLoader.loadClass(Unknown Source)
    at java.base/java.lang.Class.getDeclaredConstructors0(Native Method)
    at java.base/java.lang.Class.privateGetDeclaredConstructors(Unknown Source)
    at java.base/java.lang.Class.getConstructor0(Unknown Source)
    at java.base/java.lang.Class.getConstructor(Unknown Source)
    at org.codehaus.xfire.transport.http.HttpChannel.sendViaClient(HttpChannel.java:108)
    at org.codehaus.xfire.transport.http.HttpChannel.send(HttpChannel.java:48)
    at org.codehaus.xfire.handler.OutMessageSender.invoke(OutMessageSender.java:26)
    at org.codehaus.xfire.handler.HandlerPipeline.invoke(HandlerPipeline.java:131)
    at org.codehaus.xfire.client.Invocation.invoke(Invocation.java:79)
    at org.codehaus.xfire.client.Invocation.invoke(Invocation.java:114)
    at org.codehaus.xfire.client.Client.invoke(Client.java:336)
    at org.codehaus.xfire.client.XFireProxy.handleRequest(XFireProxy.java:77)
    at org.codehaus.xfire.client.XFireProxy.invoke(XFireProxy.java:57)
    at com.sun.proxy.$Proxy111.authenticateApplication(Unknown Source)
```

stitakis (changes requested):

build.gradle
stitakis: @renedupont it looks to me that we need to add following dependencies. I'm testing this atm.
```
    implementation group: 'commons-httpclient', name: 'commons-httpclient', version: '3.1'
    implementation(group: 'com.atlassian.security', name: 'atlassian-cookie-tools', version: '3.2.14')
```

stitakis: @renedupont my test was successful, can you please add those dependencies. This will fix the `ClassNotFoundException`. Btw, we define dependencies in all different ways, it would be time to have one standard way to define them in `gradle.properties`.
I propose to use following format:
```
implementation('com.openshift:openshift-restclient-java:9.0.1.Final')
```
It works fine for cases with exclude too: 
``` 
implementation('com.openshift:openshift-restclient-java:9.0.1.Final') {
        exclude(group: 'org.slf4j', module: 'slf4j-api:')
        exclude(group: 'org.slf4j', module: 'slf4j-log4j12')
}
```
What do you think?

renedupont: done

stitakis: excellent! Thanks!

stitakis (approved):

PR #659, opened 2020-12-23
refactor tests

renedupont:
relates to #638 

I actually started only out to get rid of the `@InjectMock` in the test but somehow ended up in refactoring more than that:

- removed in a handful of service tests classes the usage of SpringBootTest, DirtiesContext, MockBean and Autowired annotations to only use InjectMocks and Mock because it is way faster and sufficient.
- moved identical code of base- and crowd auth test config into its own auth config class and added `IODSAuthnzAdapter` mock bean and EhCache bean to it
- Replaced manual creation of MockMvc by using @AutoConfigureMockMvc in every class except the ones that require to run without spring security
- applied some Java8+ ways of coding, removed dead/unnecessary code, ...

### DIFF
--- a/src/test/java/org/opendevstack/provision/authentication/SimpleCachingGroupMembershipManagerTest.java
+++ b/src/test/java/org/opendevstack/provision/authentication/SimpleCachingGroupMembershipManagerTest.java
@@ -19,9 +19,9 @@ import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ActiveProfiles;
 
+/** Testclass for the simple caching manager */
 @SpringBootTest
 @DirtiesContext
-/** Testclass for the simple caching manager */
 @ActiveProfiles("crowd")
 public class SimpleCachingGroupMembershipManagerTest {
 
--- a/src/test/java/org/opendevstack/provision/authentication/SimpleCachingGroupMembershipManagerTest.java
+++ b/src/test/java/org/opendevstack/provision/authentication/SimpleCachingGroupMembershipManagerTest.java
@@ -19,7 +19,6 @@ import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ActiveProfiles;
 
-/** Testclass for the simple caching manager */
 @SpringBootTest
 @DirtiesContext
 @ActiveProfiles("crowd")
--- a/src/test/java/org/opendevstack/provision/authentication/crowd/ProvAppSimpleCachingGroupMembershipManagerTest.java
+++ b/src/test/java/org/opendevstack/provision/authentication/crowd/ProvAppSimpleCachingGroupMembershipManagerTest.java
@@ -29,7 +29,6 @@ import java.rmi.RemoteException;
 import java.util.List;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
-import org.mockito.Mock;
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 
@@ -42,7 +41,7 @@ public class ProvAppSimpleCachingGroupMembershipManagerTest {
 
   @MockBean private GroupManager groupManager;
 
-  @Mock private BasicCache cache;
+  @MockBean private BasicCache cache;
 
   @Test
   public void getMemberships()
--- a/src/test/java/org/opendevstack/provision/authentication/crowd/ProvAppSimpleCachingGroupMembershipManagerTest.java
+++ b/src/test/java/org/opendevstack/provision/authentication/crowd/ProvAppSimpleCachingGroupMembershipManagerTest.java
@@ -29,6 +29,7 @@ import java.rmi.RemoteException;
 import java.util.List;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.Mock;
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.test.context.junit.jupiter.SpringExtension;
 
@@ -41,7 +42,7 @@ public class ProvAppSimpleCachingGroupMembershipManagerTest {
 
   @MockBean private GroupManager groupManager;
 
-  @MockBean private BasicCache cache;
+  @Mock private BasicCache cache;
 
   @Test
   public void getMemberships()
--- a/src/test/java/org/opendevstack/provision/config/AuthSecurityTestConfig.java
+++ b/src/test/java/org/opendevstack/provision/config/AuthSecurityTestConfig.java
@@ -16,7 +16,10 @@ package org.opendevstack.provision.config;
 import static java.util.Map.entry;
 import static org.mockito.Mockito.mock;
 
+import com.atlassian.crowd.service.cache.BasicCache;
+import com.atlassian.crowd.service.cache.CacheImpl;
 import java.util.Map;
+import net.sf.ehcache.CacheManager;
 import org.opendevstack.provision.adapter.IODSAuthnzAdapter;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.cache.ehcache.EhCacheManagerFactoryBean;
@@ -59,6 +62,18 @@ public class AuthSecurityTestConfig {
   @Primary
   public IODSAuthnzAdapter iodsAuthnzAdapter() {
     return mock(IODSAuthnzAdapter.class);
+  }
+
+  @Bean
+  @Primary
+  public BasicCache getCache() {
+    return new CacheImpl(getCacheManager());
+  }
+
+  @Bean
+  @Primary
+  public CacheManager getCacheManager() {
+    return getEhCacheFactory().getObject();
   }
 
   @Bean
--- a/src/test/java/org/opendevstack/provision/config/AuthSecurityTestConfig.java
+++ b/src/test/java/org/opendevstack/provision/config/AuthSecurityTestConfig.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright 2017-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package org.opendevstack.provision.config;
+
+import static java.util.Map.entry;
+import static org.mockito.Mockito.mock;
+
+import java.util.Map;
+import org.opendevstack.provision.adapter.IODSAuthnzAdapter;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.cache.ehcache.EhCacheManagerFactoryBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+import org.springframework.security.crypto.password.PasswordEncoder;
+
+@Configuration
+public class AuthSecurityTestConfig {
+
+  public static final String TEST_USER_USERNAME = "user";
+  public static final String TEST_ADMIN_USERNAME = "admin";
+  public static final String TEST_ADMIN_EMAIL = "admin@example.com";
+  public static final String TEST_NOT_PERMISSIONED_USER_USERNAME = "not-permissioned-user";
+  public static final String TEST_VALID_CREDENTIAL = "validsecret";
+
+  @Value("${idmanager.group.opendevstack-users}")
+  protected String roleUser;
+
+  @Value("${idmanager.group.opendevstack-administrators}")
+  protected String roleAdmin;
+
+  @Bean
+  public PasswordEncoder passwordEncoder() {
+    return new BCryptPasswordEncoder();
+  }
+
+  @Bean(name = "testUsersAndRoles")
+  public Map<String, String> testUsersAndRoles() {
+    return Map.ofEntries(
+        entry(TEST_USER_USERNAME, roleUser),
+        entry(TEST_ADMIN_USERNAME, roleAdmin),
+        entry(TEST_NOT_PERMISSIONED_USER_USERNAME, "not-opendevstack-user"));
+  }
+
+  @Bean
+  @Primary
+  public IODSAuthnzAdapter iodsAuthnzAdapter() {
+    return mock(IODSAuthnzAdapter.class);
+  }
+
+  @Bean
+  @Primary
+  public EhCacheManagerFactoryBean getEhCacheFactory() {
+    EhCacheManagerFactoryBean factoryBean = new EhCacheManagerFactoryBean();
+    factoryBean.setConfigLocation(new ClassPathResource("crowd-ehcache.xml"));
+    factoryBean.setShared(false);
+    factoryBean.setAcceptExisting(true);
+    return factoryBean;
+  }
+}
--- a/src/test/java/org/opendevstack/provision/controller/DefaultControllerTest.java
+++ b/src/test/java/org/opendevstack/provision/controller/DefaultControllerTest.java
@@ -41,7 +41,6 @@ import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.result.MockMvcResultHandlers;
 import org.springframework.test.web.servlet.result.MockMvcResultMatchers;
-import org.springframework.web.context.WebApplicationContext;
 
 @SpringBootTest
 @AutoConfigureMockMvc(addFilters = false)
@@ -61,8 +60,6 @@ public class DefaultControllerTest {
   @Autowired private IBugtrackerAdapter realJiraAdapter;
 
   @Autowired private ICollaborationAdapter realConfluenceAdapter;
-
-  @Autowired private WebApplicationContext context;
 
   @Autowired private MockMvc mockMvc;
 
--- a/src/test/java/org/opendevstack/provision/controller/DefaultControllerTest.java
+++ b/src/test/java/org/opendevstack/provision/controller/DefaultControllerTest.java
@@ -22,7 +22,6 @@ import java.util.List;
 import org.hamcrest.CoreMatchers;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.mockito.Mock;
 import org.mockito.Mockito;
 import org.opendevstack.provision.adapter.IBugtrackerAdapter;
 import org.opendevstack.provision.adapter.ICollaborationAdapter;
@@ -33,6 +32,7 @@ import org.opendevstack.provision.authentication.crowd.CrowdAuthenticationManage
 import org.opendevstack.provision.model.AboutChangesData;
 import org.opendevstack.provision.services.StorageAdapter;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.security.core.context.SecurityContextHolder;
@@ -41,10 +41,10 @@ import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.web.servlet.MockMvc;
 import org.springframework.test.web.servlet.result.MockMvcResultHandlers;
 import org.springframework.test.web.servlet.result.MockMvcResultMatchers;
-import org.springframework.test.web.servlet.setup.MockMvcBuilders;
 import org.springframework.web.context.WebApplicationContext;
 
 @SpringBootTest
+@AutoConfigureMockMvc(addFilters = false)
 @ActiveProfiles({"crowd", "quickstarters"})
 public class DefaultControllerTest {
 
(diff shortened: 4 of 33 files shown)


src/test/java/org/opendevstack/provision/authentication/SimpleCachingGroupMembershipManagerTest.java
```diff
@@ -19,9 +19,9 @@
 import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ActiveProfiles;
 
+/** Testclass for the simple caching manager */
```
stitakis, line 22: This comment does not provide much help... it can be removed

renedupont: done

renedupont:
fyi: I am not sure yet why the tests are failing now after removing only one line of comment. I don't have these issues locally.

src/test/java/org/opendevstack/provision/authentication/crowd/ProvAppSimpleCachingGroupMembershipManagerTest.java
```diff
@@ -41,7 +42,7 @@
 
   @MockBean private GroupManager groupManager;
 
-  @MockBean private BasicCache cache;
+  @Mock private BasicCache cache;
```
stitakis, line 45: `@MockBean private BasisCache cache` looked ok for me... why you change it to use mockito `@Mock` annotation... 
the idea is to get rid of mockito annotation in test classes because it causes sometime confusion and troubles when mixing it with spring mockito support in same test class.

src/test/java/org/opendevstack/provision/controller/DefaultControllerTest.java
```diff
@@ -44,35 +45,31 @@
...
-@ActiveProfiles({"crowd", "utestcrowd", "quickstarters"})
+@ActiveProfiles({"crowd", "quickstarters"})
 public class DefaultControllerTest {
 
   @MockBean private IJobExecutionAdapter jobExecutionAdapter;
 
   @MockBean private StorageAdapter storageAdapter;
 
-  @MockBean private CrowdAuthenticationManager crowdAuthenticationManager;
+  @Mock private CrowdAuthenticationManager crowdAuthenticationManager;
```
stitakis, line 55: Same as above, I think we should go for `@MockBean` instead of `@Mock` to not mix mocking utilities.

src/test/java/org/opendevstack/provision/controller/E2EProjectAPIControllerTest.java
```diff
@@ -73,40 +77,35 @@
...
   public static final String BITBUCKET_PARAM_PERMISSION = "permission";
 
-  public static final String TEST_ADMIN_USERNAME = BasicAuthSecurityTestConfig.TEST_ADMIN_USERNAME;
-  public static final String TEST_VALID_CREDENTIAL =
-      BasicAuthSecurityTestConfig.TEST_VALID_CREDENTIAL;
-
   private RestClientMockHelper mockHelper;
 
-  private MockMvc mockMvc;
+  @Autowired private MockMvc mockMvc;
```
stitakis, line 108: Just to double check, does this also apply `SecurityMockMvcConfigurers.springSecurity()` to the `MockMvc` instance. See below deleted line 162

src/test/java/org/opendevstack/provision/controller/E2EProjectAPIControllerTest.java
```diff
@@ -631,13 +615,13 @@ public void testProvisionNewSimpleProjectInternal(boolean fail, boolean specialP
       return;
     }
 
-    // get the project thru its key
+    // get the project through its key
```
stitakis, line 618: comment does not provide much value, better to remove it

src/test/java/org/opendevstack/provision/controller/E2EProjectAPIControllerTest.java
```diff
@@ -795,15 +779,15 @@ public void testQuickstarterProvisionOnNewOpenProjectInclDeleteSingleComponent()
             .readValue(
                 resultProjectGetResponse.getResponse().getContentAsString(), OpenProjectData.class);
 
-    assertTrue(resultProject.getQuickstarters().size() == (currentQuickstarterSize - 1));
+    assertEquals((currentQuickstarterSize - 1), resultProject.getQuickstarters().size());
 
-    // retrieve the project thru the get endpoint - to ensure we have the right data stored
+    // retrieve the project through the get endpoint - to ensure we have the right data stored
```
stitakis, line 784: comment says exactly the same as the code express, so it does not add much value, better to remove

src/test/java/org/opendevstack/provision/services/CleanupAtlassianProjectsMT.java
```diff
@@ -20,6 +20,7 @@
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.boot.test.context.SpringBootTest.WebEnvironment;
+import org.springframework.boot.test.mock.mockito.MockBean;
 import org.springframework.core.env.Environment;
 import org.springframework.test.context.ActiveProfiles;
```
stitakis, line 26: We should add to the class name the suffix `Test`

src/test/java/org/opendevstack/provision/authentication/crowd/ProvAppSimpleCachingGroupMembershipManagerTest.java
renedupont: Hm yes I can't really remember I have to admit, but looking at the difference of ´@Mock´, ´@InjectMocks´ and ´@MockBean´ I have to say that I don't agree that we should only use ´@MockBean´. They have differences and can be potentially used for different scenarios.
One major difference for example is that `@Mock` just creates simple Mock objects which are way more performant in the test execution than ´@MockBean´. ´@MockBean´ can be used to add a Mock into the Spring application context (or replace an existing bean in there).
So I'd say, whenever a Mock does not need to be in the application context, it makes sense considering ´@Mock´ instead of ´@MockBean´. ´@InjectMocks´ creates an actual object but allows to use/inject mocks for dependencies of this class. I will need to get myself more familiar with the different scenarios and when which to use. Thus I'd rather wait with this PR. If you have any further insight please lets discuss.

src/test/java/org/opendevstack/provision/controller/E2EProjectAPIControllerTest.java
renedupont: I asked myself the same thing, and apparently the class annotation `AutoConfigureMockMvc` adds spring security if it finds it as a dependency in the project.

src/test/java/org/opendevstack/provision/services/CleanupAtlassianProjectsMT.java
renedupont: Do you know by any chance what MT in the class name stands for?

src/test/java/org/opendevstack/provision/authentication/crowd/ProvAppSimpleCachingGroupMembershipManagerTest.java
renedupont: ok after reading through https://rieckpil.de/difference-between-mock-and-mockbean-spring-boot-applications/ I think we can improve very much the test situation and performance compared to its current state. 

As a rule of thumb I'd suggest the following approach (not sure if it works everywhere yet though)
- Controller test classes get the spring boot context by the SpringBootTest and MockBean annotations
- Service test classes only have ExtendWith(MockitoExtension.class) annotation and the service the test classes tests gets InjectMocks annotation while its dependencies (collaborateurs) are annotated with Mock if necessary.

I'll try this out with the codebase of this PR.

src/test/java/org/opendevstack/provision/services/CleanupAtlassianProjectsMT.java
renedupont: looking at them it looks like it could mean 'ManualTest'

src/test/java/org/opendevstack/provision/authentication/crowd/ProvAppSimpleCachingGroupMembershipManagerTest.java
stitakis: I'd like to stick first to `@MockBean` and improve it later.
To avoid to delay much the review and merge of this PR, I'll propose to do the "situation and performance improvement in a 2nd change after finalizing and merging this one.
Sounds good?

renedupont: yes ok, I already changed it for a handful smaller tests class were it was applicable and easy to fix (see latest commit: https://github.com/opendevstack/ods-provisioning-app/pull/659/commits/e46af5e24b447f75a13dab40e03264dc31b152e3). I think I'll stop it here then until we merged.

src/test/java/org/opendevstack/provision/controller/E2EProjectAPIControllerTest.java
renedupont: done

src/test/java/org/opendevstack/provision/controller/E2EProjectAPIControllerTest.java
stitakis: Since Spring Security add web filters, I tested the class with `@AutoConfigureMockMvc(addFilters = false)` with the assumption that it will fail. This was the case.
Additionally I did a manual debug to verify that the security auth annotation logic is called. It looks fine to use `@AutoConfigureMockMvc`.

src/test/java/org/opendevstack/provision/controller/DefaultControllerTest.java
```diff
@@ -44,35 +45,31 @@
...
   @BeforeEach
   public void setUp() {
-
-    mockMvc =
-        MockMvcBuilders.webAppContextSetup(context)
-            // .apply(SecurityMockMvcConfigurers.springSecurity())
-            .build();
-
+    // mockMvc without spring security
+    mockMvc = MockMvcBuilders.webAppContextSetup(context).build();
```
stitakis, line 72: This test actually needs spring security... I run it with `@AutoConfigureMockMvc` and all test passed. Let us add the annotation `@AutoConfigureMockMvc` instead of line 72.

src/test/java/org/opendevstack/provision/controller/ProjectApiControllerTest.java
```diff
@@ -249,7 +244,7 @@ public void addProjectWithOC() throws Exception {
                     .andExpect(MockMvcResultMatchers.status().isOk())
                     .andDo(MockMvcResultHandlers.print());
 
-                verifyAddProjectAdapterCalls(times(1), times(confluenceAdapterEnabled ? 1 : 0));
```
stitakis, line 252: This line is actually ok. I realized that the method `verifyAddProjectAdapterCalls` has a bug. 
To fix it, the folllowing code needs to be added to the method otherwise it doesn't make sense to pass the 2nd variable:
```
 Mockito.verify(confluenceAdapter, confluenceTimes)
            .createCollaborationSpaceForODSProject(isNotNull());
```

src/test/java/org/opendevstack/provision/controller/DefaultControllerTest.java
stitakis, line 72: We could use `@AutoConfigureMockMvc(addFilters = false)` instead of the line above. In this case we need to add `@Autowired` to `private MockMvc mockMvc;`

src/test/java/org/opendevstack/provision/services/CleanupAtlassianProjectsMT.java
stitakis: then let us add as suffix `ManualTest`. The name will be then more expressive.

stitakis (changes requested):

src/test/java/org/opendevstack/provision/controller/DefaultControllerTest.java
renedupont: ok in this case I changed it back for you for now ;)

src/test/java/org/opendevstack/provision/services/CleanupAtlassianProjectsMT.java
renedupont: done

src/test/java/org/opendevstack/provision/controller/ProjectApiControllerTest.java
renedupont: done

src/test/java/org/opendevstack/provision/controller/DefaultControllerTest.java
renedupont: done

src/test/java/org/opendevstack/provision/services/MailAdapterTest.java
```diff
@@ -14,46 +14,41 @@
...
 import org.opendevstack.provision.authentication.crowd.CrowdAuthenticationManager;
 import org.opendevstack.provision.model.OpenProjectData;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.context.SpringBootTest;
 import org.springframework.mail.javamail.JavaMailSender;
-import org.springframework.test.annotation.DirtiesContext;
 import org.springframework.test.context.ActiveProfiles;
 
 @SpringBootTest
-@DirtiesContext
```
stitakis, line 37: can you please add `@DirtiesContext` again?

src/test/java/org/opendevstack/provision/services/MailAdapterTest.java
```diff
@@ -14,46 +14,41 @@
...
 @ActiveProfiles("crowd")
 public class MailAdapterTest {
-  @Mock JavaMailSender mailSender;
 
-  @InjectMocks @Autowired MailAdapter mailAdapter;
+  @Autowired private MailAdapter mailAdapter;
 
-  @Autowired CrowdAuthenticationManager manager;
+  @Mock private CrowdAuthenticationManager crowdAuthenticationManager;
```
stitakis, line 39: as discussed, please replace `@Mock` in favour of `@MockBean`

src/test/java/org/opendevstack/provision/services/MailAdapterTest.java
```diff
@@ -14,46 +14,41 @@
...
+
+  @Mock private JavaMailSender mailSender;
```
stitakis, line 41: same as last comment

stitakis:
@renedupont fyi... `MailAdapterTest` is failing in my dev env when I run all tests together. Something related to a missing dependency. I'm trying to figure out, what is the root of the problem.

src/test/java/org/opendevstack/provision/services/MailAdapterTest.java
renedupont: Actually, that one has to stay as Mock, otherwise you will get 'Bean must not be empty' exception and it has been Mock already before me changing anything.

src/test/java/org/opendevstack/provision/config/AuthSecurityTestConfig.java
```diff
@@ -0,0 +1,88 @@
...
+
+  @Bean
+  @Primary
+  public EhCacheManagerFactoryBean getEhCacheFactory() {
+    EhCacheManagerFactoryBean factoryBean = new EhCacheManagerFactoryBean();
+    factoryBean.setConfigLocation(new ClassPathResource("crowd-ehcache.xml"));
+    factoryBean.setShared(false);
+    factoryBean.setAcceptExisting(true);
+    return factoryBean;
+  }
```
stitakis, line 87: I found the problem that was causing the tests to fail in my dev env. Actually the 3 beans defined above `BasicCache` and `CacheManager` and `EhCacheManagerFactoryBean` are duplicated for the profile `crowd`. The other duplicates are in the class `CrowdSecurityConfiguration', which is not part of the test classes.
After commenting out the beans above my tests run successful in my dev env. 
I propose we remove the beans above to avoid bean duplication for the profile `crowd`. 

As ref: this is the stacktrace that points to the bean duplication:
```
2021-01-20 16:08:04 ERROR o.s.test.context.TestContextManager - Caught exception while allowing TestExecutionListener [org.springframework.test.context.web.ServletTestExecutionListener@775de577] to prepare test instance [
org.opendevstack.provision.services.MailAdapterTest@1587a71]
    java.lang.IllegalStateException: Failed to load ApplicationContext
        at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:132)
        at org.springframework.test.context.support.DefaultTestContext.getApplicationContext(DefaultTestContext.java:123)
        at org.springframework.test.context.web.ServletTestExecutionListener.setUpRequestContextIfNecessary(ServletTestExecutionListener.java:190)
        at org.springframework.test.context.web.ServletTestExecutionListener.prepareTestInstance(ServletTestExecutionListener.java:132)
...
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'getCache' defined in class path resource [org/opendevstack/provision/authentication/crowd/CrowdSecurityConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.atlassian.crowd.service.cache.BasicCache]: Factory method 'getCache' threw exception; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'getCacheManager' defined in class path resource [org/opendevstack/provision/authentication/crowd/CrowdSecurityConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [net.sf.ehcache.CacheManager]: Factory method 'getCacheManager' threw exception; nested exception is org
```

src/test/java/org/opendevstack/provision/services/MailAdapterTest.java
stitakis: I see... thanks!

src/test/java/org/opendevstack/provision/services/MailAdapterTest.java
stitakis: I see, the last change removed the spring test annotation for this test. Since `MailAdapter` is annotated with `@Service` and for that reason is a spring service is wiser to test it with the spring test annotation to benefit and test the DI configuration. 
However, we are having issues in getting the correct spring test configuration, which is delaying us from merging this refactoring. 
Since this refactoring improve a lot the overall tests and as I do not want to delay the merge, I propose to create 2 new issues. One to turn back the configuration of MailAdapterTest to use spring annotations. 2nd one to improve the spring test configuration which will give us performance improvements. Sound good?

src/test/java/org/opendevstack/provision/services/MailAdapterTest.java
```diff
@@ -14,70 +14,54 @@
...
 
-  @MockBean private CrowdAuthenticationManager crowdAuthenticationManager;
+  @Mock private CrowdAuthenticationManager crowdAuthenticationManager;
 
   @Mock private JavaMailSender mailSender;
 
   @BeforeEach
   public void setUp() {
+    mailAdapter = new MailAdapter(mailSender);
     mailAdapter.manager = crowdAuthenticationManager;
```
stitakis, line 43: Problem with this direct assignment to internal member of class is:
1. breaks historical and wide adopted java conventions of doing this with a setter. So devs do not expect this kind of implementation.  
2. I believe (not 100% sure) bytecode enhancer tools like asm decorates code based on getter and setter (due to limitation in the java reflextion functionality) Such tools are used by testing frameworks to instrument the code. If this rule is not followed then the tool might not be able to work as expected, which could lead to unexpected behaviour.
  
Therefore, can you please create a setter method for this?

renedupont: Hey! yes I did not change it because it did not yet wanna touch the MailAdapter class. The best way here would be to also inject this manager service via constructor injection as it is done with mailSender. I can do this later.

stitakis: ok, let us improve the `MailAdapter` in a later change.

stitakis:
@renedupont this unit test is failing
``` 
MailAdapterTest > notifyUsersAboutProjectMailEnabled() FAILED
    org.mockito.exceptions.verification.WantedButNotInvoked at MailAdapterTest.java:55
2021-01-20 21:49:50 INFO  o.s.s.c.ThreadPoolTaskExecutor - Shutting down ExecutorService 'applicationTaskExecutor'
2021-01-20 21:49:50 INFO  o.s.c.e.EhCacheManagerFactoryBean - Shutting down EhCache CacheManager
```

renedupont:
I know, it is not failing locally, therefore I have a hard time fixing it, but I have an idea and will try to fix it.

src/test/java/org/opendevstack/provision/services/MailAdapterTest.java
renedupont: Could you explain why you think it is wiser to test services with the spring test annotation? In my opinion, in services we often only want to unit test the business logic in isolation here, which can be independent of the Spring Framework. 
I see the use case for the spring application context rather when testing the integration of multiple classes or for example the web layer in isolation with a mocked servlet environment.

stitakis: the additionally benefit of testing with spring test configuration is that the spring configuration logic and how the service is built up is also tested. Otherwise that logic is not tested.
While this goes beyond of the scope of a "unit test", it has the advantage that is a bit closer test scenario to how the service is built up later in production. Make sense?

renedupont: hm I'm not quite sure about this, even the spring doc mentions in [its part about unit testing](https://docs.spring.io/spring-framework/docs/current/reference/html/testing.html#unit-testing):

> Dependency injection should make your code less dependent on the container than it would be with traditional Java EE development. The POJOs that make up your application should be testable in JUnit or TestNG tests, with objects instantiated by using the new operator, without Spring or any other container. You can use mock objects (in conjunction with other valuable testing techniques) to test your code in isolation. If you follow the architecture recommendations for Spring, the resulting clean layering and componentization of your codebase facilitate easier unit testing. **For example, you can test service layer objects by stubbing or mocking DAO or repository interfaces**, without needing to access persistent data while running unit tests.

Testing the spring context is something they mention in the [part about integration test](https://docs.spring.io/spring-framework/docs/current/reference/html/testing.html#integration-testing-overview):

> It is important to be able to perform some integration testing without requiring deployment to your application server or connecting to other enterprise infrastructure. Doing so lets you test things such as:
> 
> -   **The correct wiring of your Spring IoC container contexts.**
> ...

but anyway, I'll put it this test class back to a state where it worked and we can fix this situation later on.

stitakis (approved):
LGTM!

PR #660, opened 2021-01-05
raise version of changelog enforcer to v2 and add custom error message

renedupont:
relates to https://github.com/opendevstack/ods-core/issues/891#issuecomment-747923351


michaelsauter (approved):
Ace 👌

PR #661, opened 2021-01-06
Update changelog-enforcer.yml

renedupont:
In the changelog enforcer action I set the branches on pull request to master so that it only runs for PRs against master.


renedupont:
was closed because master is not enough, we have release branches that are also applicable and steady growing in numbers.

PR #669, opened 2021-01-28
add form based auth in spa

### LINKED ISSUE #637: Handle form based auth in SPA
On 401 responses a login form should be displayed. As a first step it will send the form data in the Basic-Auth Header via GET to /appinfo/about

netzartist:
Fixes #637 

Feature: When API requests answer with a 401 return code, display a login page which sends login credentials to an auth URL (see `environments/*)` via basic auth to receive a (new) JSESSIONID cookie

Expanded `client/README` with config instruction for SSL localhost setup 

Open question: Is an explicit logout button necessary?


stitakis:
@netzartist yes, from security point of view a logout button is required to give the user a way to terminate a session.
To logout a http POST request needs to be sent to this endpoint: `/logout`

stitakis:
@netzartist does this fix is compatible with SSO? Does it falls back to basic auth if SSO is not working?

netzartist:
@stitakis How could the app know that SSO is not available, is there an endpoint which I can ping to get a `true|false` flag?

stitakis:
@netzartist I've changed the title of the issue to reflect better the solution

stitakis:
@renedupont because I have done some changes I have removed myself from the list of reviews, can you please review the backend changes?

stitakis:
@netzartist from backend point of view, this is ready for PR review

src/main/java/org/opendevstack/provision/authentication/crowd/CrowdSecurityConfiguration.java
```diff
@@ -232,11 +222,42 @@ public void onAuthenticationSuccess(
...
 
-  /**
-   * define an failure handler in case of an authentication failure
-   *
-   * @return
-   */
+  @NotNull
+  public SimpleUrlAuthenticationSuccessHandler createAuthSuccessHandlerThatReturnsOK() {
+    SimpleUrlAuthenticationSuccessHandler handler =
+        new SimpleUrlAuthenticationSuccessHandler() {
```
renedupont, line 228: I sense that this could be expressed as a lambda expression

src/main/java/org/opendevstack/provision/authentication/crowd/CrowdSecurityConfiguration.java
```diff
@@ -110,13 +116,22 @@ protected void configure(HttpSecurity http) throws Exception {
...
         .antMatchers(
-            "/", "/fragments/**", "/webjars/**", "/js/**", "/json/**", "/favicon.ico", "/login")
+            "/",
+            "/fragments/**",
+            "/webjars/**",
+            "/js/**",
+            "/json/**",
+            "/favicon.ico",
+            "/login",
+            "/nfe/**")
```
renedupont, line 134: what does nfe stand for? this abbreviation is often used but I don't know what it means.

src/main/java/org/opendevstack/provision/authentication/crowd/CrowdSecurityConfiguration.java
```diff
@@ -232,11 +222,42 @@ public void onAuthenticationSuccess(
...
+  public static void logSuccessAuthentication(Authentication authentication) {
+    try {
+      String username = null;
+
+      if (authentication.getPrincipal() instanceof CrowdUserDetails) {
+        CrowdUserDetails userDetails = (CrowdUserDetails) authentication.getPrincipal();
+        username = userDetails.getUsername();
+      }
+
+      logger.info("Successful authentication [username=" + username + "]");
```
renedupont, line 254: shouldnt this be in the if clause above? otherwise username is null

src/main/java/org/opendevstack/provision/authentication/crowd/CrowdSecurityConfiguration.java
renedupont: Ok got now that it stands for NewFrontEnd but I'd still write it out like this otherwise this information gets lost.

stitakis: NFE=new front end -> we are planning to rename this in a different change

src/main/java/org/opendevstack/provision/authentication/crowd/CrowdSecurityConfiguration.java
stitakis: yes, it could :-)

netzartist:
@cschweikert I'd be happy to have you review the frontend changes. We've reduced the scope to login functionality, logout logic will be tackled further in another PR.

client/src/app/modules/login/components/login.component.html
```diff
@@ -0,0 +1,66 @@
...
+    <app-loading-indicator></app-loading-indicator>
+  </ng-container>
+
+  <form *ngIf="!isLoading" [formGroup]="form" (ngSubmit)="intendFormSubmit()" novalidate id="form">
+  <div class="box">
+    <div class="page__header">
+      <h3 class="mat-h3">Login</h3>
+    </div>
+    <div class="page__content">
+      <div class="row">
```
cschweikert, line 12: I'm wondering, if we really need the wrapping divs with class ``row``. Would it also work like this:
```
<div class="page__content col">
  <div class="col">...username-related stuff...</div>
  <div class="col">...password-related stuff...</div>
</div>
```

client/src/app/app.component.ts
```diff
@@ -71,17 +98,13 @@ export class AppComponent implements OnInit {
...
     this.projectService
       .getAllProjects()
-      .pipe(
-        catchError(() => {
-          this.isError = true;
-          this.isLoading = false;
-          return EMPTY;
-        })
-      )
+      .pipe(catchError(() => EMPTY))
```
cschweikert, line 101: maybe I'm wrong, but doesn't the ``EMPTY`` lead to no execution of the function handed over with ``.subscribe(...)``?
I tried it with ``const subscribe = EMPTY.subscribe(() => console.log('Hi'));`` and that didn't say "Hi".
Technically the ``EMPTY`` triggers the complete handler right away, which is not provided here.

client/src/app/app.component.html
```diff
@@ -4,7 +4,7 @@
       <div class="logo">
         <h1>ODS Provisioning</h1>
       </div>
-      <div class="user-area" [hidden]="router.url.indexOf('/logout') !== -1">
+      <div class="user-area" [hidden]>
```
cschweikert, line 7: Why does ``hidden`` have ``[...]`` around and will it then be always hidden?

cschweikert (changes requested):

client/src/app/modules/error-handler/http-request-interceptor.ts
```diff
@@ -40,10 +40,7 @@ export class HttpRequestInterceptor implements HttpInterceptor {
 
     return next.handle(req).pipe(
       catchError((error: HttpErrorResponse) => {
-        if (error.status === 401) {
```
cschweikert: I guess the purpose of this if statement was to differ between authentication related errors and any other kind of error. If the if statement is gone, a navigation change will always be triggered.

renedupont (approved):
I only reviewed backend

client/src/app/modules/login/components/login.component.ts
```diff
@@ -0,0 +1,70 @@
...
+  private initializeFormGroup(): void {
+    this.form = this.formBuilder.group({
+      username: ['', [Validators.required]],
+      password: ['', [Validators.required]]
+    });
+  }
+
+  private initiateLogin(): void {
+    this.isLoading = true;
+    this.authenticationService
```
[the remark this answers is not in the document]
netzartist: Doesn't work this way as the `Validators.required` will complain about empty fields then. Placed a `this.form.reset()` in the subscription success callback.

client/src/app/modules/login/components/login.component.html
netzartist: Actually it works also without the `col` at all as they are just 2 form fields below each other.

client/src/app/app.component.html
netzartist: Oops, of course only `hidden` is needed here.

client/src/app/modules/error-handler/http-request-interceptor.ts
netzartist: Re-added the if clause.

client/src/app/modules/login/components/login.component.ts
```diff
@@ -0,0 +1,70 @@
...
+import { first } from 'rxjs/operators';
+import { Router } from '@angular/router';
+
+@Component({
+  selector: 'app-login',
+  templateUrl: './login.component.html',
+  styleUrls: ['./login.component.scss']
+})
+export class LoginComponent extends FormBaseComponent implements OnInit, OnDestroy {
+  destroy$ = new Subject<boolean>();
```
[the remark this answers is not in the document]
netzartist: In order to unsubscribe from all observables. My habit is basically using `takeUntil` like mentioned in no. 3 in [this approach](https://medium.com/angular-in-depth/the-best-way-to-unsubscribe-rxjs-observable-in-the-angular-applications-d8f9aa42f6a0), but maybe there's a better way meanwhile...

client/src/app/modules/logout/components/logout.component.ts
```diff
@@ -1,19 +1,23 @@
...
 import { BrowserService } from '../../browser/services/browser.service';
 import { Subject } from 'rxjs';
+import { AuthenticationService } from '../../authentication/services/authentication.service';
 
 @Component({
   selector: 'app-logout',
   templateUrl: './logout.component.html'
 })
 export class LogoutComponent implements OnInit, OnDestroy {
   destroy$ = new Subject<boolean>();
```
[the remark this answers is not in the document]
netzartist: See [other comment](https://github.com/opendevstack/ods-provisioning-app/pull/669#discussion_r585542362)

client/src/app/modules/login/components/login.component.ts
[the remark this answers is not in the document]
cschweikert: I can't see where it is actually used. Maybe it is not needed anymore since there is a ``.pipe(first())`` instead (see also no. 4).

client/src/app/modules/logout/components/logout.component.ts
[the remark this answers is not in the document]
cschweikert: ``this.destroy$`` doesn't seem to be used. Maybe use ``.pipe(first())`` within ``ngOnInit()``.

client/src/app/app.component.ts
netzartist: Good comment! But ... hm ... `EMPTY` docu says `Creates an Observable that emits no items to the Observer and immediately emits a complete notification`. In my case when there's an error then the observable should "stop" and the http interceptor kicks in, that's why the component isn't involved anymore and I found it safe that `catchError` emits "complete". Ok for you to further talk about this outside of this PR as there was no changes introduced to this piece of code?

client/src/app/modules/logout/components/logout.component.ts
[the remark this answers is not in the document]
netzartist: Refactored it

client/src/app/modules/login/components/login.component.ts
```diff
@@ -0,0 +1,70 @@
...
+      .pipe(first())
+      .subscribe(
+        () => {
+          this.renderer.removeClass(document.body, 'status-user-auth');
+          this.router.navigateByUrl('/');
+        },
+        () => {
+          this.isLoading = false;
+          this.isLoginError = true;
+          return EMPTY;
```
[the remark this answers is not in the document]
netzartist: No, removed it.

client/src/app/modules/login/components/login.component.ts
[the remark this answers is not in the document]
netzartist: Refactored it.

client/src/app/app.component.ts
netzartist: Added gone-lost boolean variables in the `catchError` callback.

cschweikert (approved):

renedupont (approved):

PR #671, opened 2021-02-11
fixes project specific cd user not permissioned on provision new component repository

### LINKED ISSUE #670: Project specific CD user not permissioned on provision of new component repository
Finding together with @clemensutschig :
When a new component is provisioned, the bitbucket technical user is still set with write permissions to that repository here:
https://github.com/opendevstack/ods-provisioning-app/blob/27d561e2f6b82cfae75c34398b3d08f257296e74/src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java#L660

This allows access to the repos with the global cd_user even if a project specific user is set and thus is a security issue.

Maybe we chan do it somehow like below to use the project specific cdUser instead of the global one:
https://github.com/opendevstack/ods-provisioning-app/blob/27d561e2f6b82cfae75c34398b3d08f257296e74/src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java#L254-L274

stitakis:
fixes #670

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Fixed
 
+- Project specific CD user not permissioned on provision of new component repository ([#670](https://github.com/opendevstack/ods-provisioning-app/issues/670))
 - Show an error message when the component id does not match the regex expression ([#624](https://github.com/opendevstack/ods-provisioning-app/issues/624))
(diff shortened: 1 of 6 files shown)


michaelsauter (approved):
Change looks good to me.

As discussed we should have another look at this, but as a quick bandaid, without restructuring the permission approach, this works.

PR #676, opened 2021-03-05
Handle logout in SPA

### LINKED ISSUE #675: Handle logout in SPA
Following up on #637, the user should be able to logout from the provisioning app. A `/logout` endpoint is already available.

netzartist:
Fixes #675


client/src/app/app.component.ts
```diff
@@ -99,10 +105,14 @@ export class AppComponent implements OnInit {
     this.projectService
       .getAllProjects()
       .pipe(
-        catchError(() => {
+        catchError((response: HttpErrorResponse) => {
+          if (response.error.status === 401) {
```
cschweikert, line 109: When is the interceptor jumping in? Isn't it intercepting even before this line is executed?

client/src/app/app.component.ts
```diff
@@ -99,10 +105,14 @@ export class AppComponent implements OnInit {
...
+            // interceptor takes control and redirects to login
+            return EMPTY;
+          }
+          // show generic error page
           this.isLoading = false;
           this.isError = true;
-          return EMPTY;
```
cschweikert, line 105: Does it work without returning (another) observable?

netzartist: No, returning an `EMPTY` observable is necessary ;)

client/src/app/app.component.ts
netzartist: Right, I was a bit on the wrong track due to a bug in the interceptor which is fixed now. In 401 cases the interceptor needs to return an `EMPTY` observable in order to let the app component render the login page through the router outlet in this case.

client/src/app/app.component.ts
```diff
@@ -55,24 +56,8 @@ export class AppComponent implements OnInit {
...
-  }
-
-  showLogoutButton() {
-    this.router.events.subscribe(event => {
-      if (event instanceof NavigationStart && (event.url === 'login' || event.url === 'logout')) {
-        return false;
-      }
-      return true;
-    });
+    this.checkShowLogoutButton();
```
cschweikert, line 60: Maybe just a minor thing: a few lines above we are already looking at ``router.events``. Maybe the code can be joined, e.g. in the following way:
```
this.router.events.subscribe(event => {
  if (event instanceof NavigationStart) {
    if (event.url === '/') {
      this.loadAllProjects();
    } else if (event.url === 'login' || event.url === 'logout') {
      this.showLogoutButton = false;
    }
    this.isLoading = false;
  }
});
```

BTW: I'm wondering, if ``this.isLoading = false;`` should not be skipped when calling ``this.loadAllProjects();``.

client/src/app/app.component.ts
```diff
@@ -89,10 +74,29 @@ export class AppComponent implements OnInit {
...
+  private checkShowLogoutButton() {
+    return this.router.events.subscribe(event => {
+      if (event instanceof NavigationStart && (event.url === '/login' || event.url === '/logout')) {
+        this.showLogoutButton = false;
+      }
+    });
+  }
+
+  private setSsoMode() {
+    this.authenticationService.sso = true;
```
cschweikert, line 86: just a thought: Would it make sense to move the whole sso detection (``this.activatedRoute.queryParams.subscribe(...)`` or maybe just the ``setSsoMode()`` function including the ``this.router.navigate(...)`` into the ``AuthenticationService``?

cschweikert (approved):

client/src/app/app.component.ts
netzartist: Good point! And `this.isLoading` should be in the `else if` part, as `this.loadAllProjects()` sets this on its own.

client/src/app/app.component.ts
netzartist: Oh yeah, that's a better separation of concerns.

client/src/app/app.component.ts
```diff
@@ -40,35 +42,20 @@ export class AppComponent implements OnInit {
...
-        this.setSsoMode();
-      }
-    });
+    this.authenticationService.checkAndSetSsoMode();
 
     this.router.events.subscribe(event => {
       if (event instanceof NavigationStart) {
         if (event.url === '/') {
           this.loadAllProjects();
+        } else if (event instanceof NavigationStart && (event.url === '/login' || event.url === '/logout')) {
```
cschweikert, line 51: ``event instanceof NavigationStart && `` can be skipped here. It is checked earlier

client/src/app/app.component.ts
```diff
@@ -40,35 +42,20 @@ export class AppComponent implements OnInit {
   }
 
   ngOnInit() {
-    this.activatedRoute.queryParams.subscribe(params => {
```
cschweikert, line 43: ``this.activatedRoute`` can also be removed from the constructor

client/src/app/app.component.ts
netzartist: Done

client/src/app/app.component.ts
netzartist: Done

cschweikert (approved):

PR #679, opened 2021-03-25
Add swagger-ui support

stitakis:


renedupont:
Hey! Should this not be secured by spring security as well?

stitakis:
It is secured. If enabled, the swagger ui, which is meant in ProvApp to be used in a development environment, can't be accessed by an authenticated users.

renedupont:
ok I see. I guess I'd suggest to add to our improvement list the proper usage of spring profiles instead of this conditional approach. But for now seems ok, if you fix the conflicting file I can approve

renedupont (approved):

PR #700, opened 2021-11-17
Fix problem assigning admin permissions to bitbucket repositories

jafarre-viewnext:
Fix admin permission assignment for new coponents


src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
```diff
@@ -658,7 +659,9 @@ protected RepositoryData callCreateRepoApi(String projectKey, Repository repo)
                   + " - no response from endpoint, please check logs",
               repo.getName(), projectKey));
     }
-    setRepositoryAdminPermissions(data, projectKey, ID_GROUPS, repo.getUserGroup());
+    if (project.isSpecialPermissionSet()) {
```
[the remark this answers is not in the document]
stitakis: after discussing this we introduce following changes:
if special permission was set
- admin permission to the repo admin group
- admin permission to the repo admin user
if special permission was not set
- admin permission to the repository creator
in any case we will set
- write permission to repo user group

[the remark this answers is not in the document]
jafarre-viewnext: After a deeper review, we find:

- If special permissions are set, permissions are correctly applied at project level and inherited at repo level.
- If special permissions are not set, nobody gets admin permissions at project level and the default user group gets write permissions, as desired. Providing admin permissions to the repository creator does not seem possible. It may even be a technical user.

So, everything is fine when special permissions are set.
I will create a table with the currently applied permissions for further discussion about the desired behaviour when special permissions are not set.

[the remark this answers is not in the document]
jafarre-viewnext: This conversation is around outdated code. Resolving the conversation.

jafarre-viewnext:
The problem is that, when a new repository is created, we are assigning admin permissions to a generic user group.
This assignment has to be removed, but we need to validate that the resulting set of permissions is appropriate.
@metmajer @clemensutschig this is not a technical decision and we would need you to validate the assigned permissions. Can you please review it?
The list of assigned permissions follows.

Currently assigned permissions when the special permission set is specified on project creation:

| Type    | Who?                                        | Level    | Permission        |
| ------- | --------------------------------- | -------- | ----------------- |
| Group | ${global.keyuser.role.name}     | project | ADMIN              |
| Group | project.projectAdminGroup     | project | ADMIN              |
| Group | project.projectUserGroup        | project | WRITE                |
| Group | project.projectReadonlyGroup | project | READ                 |
| User | project.cdUser     | project | WRITE              |
| User | project.cdUser     | repo | READ              |
| **_Group_** | **_${bitbucket.default.user.group}_**     | **_repo_** | **_ADMIN_**              |
| User | ${bitbucket.technical.user}     | project | WRITE              |

Currently assigned permissions when the special permission set is **not** specified on project creation:

| Type    | Who?                                        | Level    | Permission        |
| ------- | --------------------------------- | -------- | ----------------- |
| Group | ${bitbucket.default.user.group} | project | WRITE                |
| Group | ${idmanager.group.opendevstack-users} | project | READ                 |
| User | project.cdUser     | project | WRITE              |
| **_Group_** | **_${bitbucket.default.user.group}_**     | **_repo_** | **_ADMIN_**              |
| User | ${bitbucket.technical.user}     | project | WRITE              |

Additionally, whenever a specific CD User is specified on project creation, this user gets read permissions in all repositories in scmGlobalProperties.readableRepos (such as opendevstack/ods-jenkins-shared-library and opendevstack/ods-quickstarters).

The highlighted row is the one detected as giving admin permissions to a generic user group and has to be removed.
If we do that, in case the special permission set is **not** specified, then the respository would not get any specific admin group or user. The only administrators would be the global bitbucket administrators. Is this fine?

jafarre-viewnext:
Removed the admin permissions at repo level.
This is the resuting permissions, that need to be validated:

Assigned permissions after the fix, when the special permission set is specified on project creation:

| Type    | Who?                                        | Level    | Permission        |
| ------- | --------------------------------- | -------- | ----------------- |
| Group | ${global.keyuser.role.name}     | project | ADMIN              |
| Group | project.projectAdminGroup     | project | ADMIN              |
| Group | project.projectUserGroup        | project | WRITE                |
| Group | project.projectReadonlyGroup | project | READ                 |
| User | project.cdUser (default: ${bitbucket.technical.user}) | project | WRITE              |
| User | ${bitbucket.technical.user}     | repo | WRITE              |

Assigned permissions after the fix, when the special permission set is **not** specified on project creation:

| Type    | Who?                                        | Level    | Permission        |
| ------- | --------------------------------- | -------- | ----------------- |
| Group | ${bitbucket.default.user.group} | project | WRITE                |
| Group | ${idmanager.group.opendevstack-users} | project | READ                 |
| User | project.cdUser (default: ${bitbucket.technical.user}) | project | WRITE              |
| User | ${bitbucket.technical.user}     | repo | WRITE              |

Additionally, whenever a specific CD User is specified on project creation, this user gets read permissions in all repositories in scmGlobalProperties.readableRepos (such as opendevstack/ods-jenkins-shared-library and opendevstack/ods-quickstarters).

jafarre-viewnext:
After discussion, the decissions taken are:

- If special permissions set has been specified, we will not assign a specific admin group to the repository. It will inherit the admin group from the project.
- For the case that special permissions set has not been specified, we create a new _bitbucket.default.admin.group_ config property and assign admin permissions to it.
- If _bitbucket.default.admin.group_ is not specified, it will default to _bitbucket.default.user.group_.
- In the unlikely case that _bitbucket.default.user.group_ is not specified, we will not assign specific admin permissions to the repository. Only the global Bitbucket administrators will have admin access to the repo.
- Auxiliary repositories (occonfig-artifacts and design) will get the same permissions as the regular repositories.

The final permission tables follow.

Assigned permissions after the fix, when the special permission set is specified on project creation:

| Type    | Who?                                        | Level    | Permission        |
| ------- | --------------------------------- | -------- | ----------------- |
| Group | ${global.keyuser.role.name}     | project | ADMIN              |
| Group | project.projectAdminGroup     | project | ADMIN              |
| Group | project.projectUserGroup        | project | WRITE                |
| Group | project.projectReadonlyGroup | project | READ                 |
| User | project.cdUser (default: ${bitbucket.technical.user}) | project | WRITE              |
| User | ${bitbucket.technical.user}     | repo | WRITE              |

Assigned permissions after the fix, when the special permission set is **not** specified on project creation:

| Type    | Who?                                        | Level    | Permission        |
| ------- | --------------------------------- | -------- | ----------------- |
| Group | ${bitbucket.default.user.group} | project | WRITE                |
| Group | ${idmanager.group.opendevstack-users} | project | READ                 |
| User | project.cdUser (default: ${bitbucket.technical.user}) | project | WRITE              |
| Group | ${bitbucket.default.admin.group} (default: ${bitbucket.default.user.group} | repo| ADMIN                 |
| User | ${bitbucket.technical.user}     | repo | WRITE              |

stitakis:
@jafarre-viewnext happy to review your changes once you have implemented the rules specified in your last comment

jafarre-viewnext:
> @jafarre-viewnext happy to review your changes once you have implemented the rules specified in your last comment

Thank you, @stitakis ! I will set the PR as ready for review as soon as I have properly tested the changes.

jafarre-viewnext:
Some notes about the implementation:

The new property has been added to the application.properties as follows:
`bitbucket.default.admin.group=${bitbucket.default.user.group}`
The code just honors this property value. This is enough to maintain backwards compatibility when the new property is not present in the config map.
However, this also allows users to disable the admin group by specifying the following in the config map:
`bitbucket.default.admin.group=`

There was a problem when assigning permissions on project creation. If the same group was specified as having, for instance, write and admin permissions, The admin permissions were overwritten by the write permissions.
This was actually happenning with the default user group, but the problem was hidden, because this group was getting admin permissions at repo level.
It has been fixed by reversing the order in which permissions are assigned.

jafarre-viewnext:
@stitakis ready for review. Thanks in advance!

stitakis (changes requested):
I think is time to document the implemented rules.
Proper documentation for this is missing and causes recurring queries on how the provision app behaves.
Can you please extends https://github.com/opendevstack/ods-provisioning-app/blob/master/docs/modules/provisioning-app/pages/index.adoc to capture the new implemented rules?

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
```diff
@@ -651,8 +651,9 @@ protected BitbucketProjectData callCreateProjectApi(OpenProjectData project) thr
     return projectData;
   }
 
-  protected RepositoryData callCreateRepoApi(String projectKey, Repository repo)
+  protected RepositoryData callCreateRepoApi(OpenProjectData project, Repository repo)
```
stitakis, line 654: A small improvement, but I think it improves readability from the method caller point of view. I would prefer here to  additionally pass the boolean `isSpecialPermissionSet` instead of passing the object `OpenProjectData`. The method signature would look like this:
``` 
protected RepositoryData callCreateRepoApi(String projectKey, boolean specialPermissionSet, Repository repo)
``` 
Can you please change it?

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
```diff
@@ -664,20 +665,13 @@ protected RepositoryData callCreateRepoApi(String projectKey, Repository repo)
...
-          data.getSlug());
-      setRepositoryPermissions(
-          data.getSlug(),
-          projectKey,
-          ID_GROUPS,
-          openDevStackUsersGroupName,
-          REPOSITORY_PERMISSIONS.REPO_WRITE);
+    if (!project.isSpecialPermissionSet()) {
+      String adminGroup = repo.getAdminGroup();
+      if (adminGroup != null && !adminGroup.isEmpty()) {
```
stitakis, line 670: It looks to me that this logic can be moved to the method callCreateProjectApi because the permission should be inherit from the bitbucket project instead to be set explicity in at repository level. What do you think?

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
```diff
@@ -477,14 +477,14 @@ public static boolean containsGroup(JsonNode json, String group) {
           repo.setAdminGroup(project.getProjectAdminGroup());
           repo.setUserGroup(project.getProjectUserGroup());
         } else {
-          repo.setAdminGroup(defaultUserGroup);
+          repo.setAdminGroup(defaultAdminGroup);
```
stitakis, line 480: This seems to not be covered with a unit test. Can you please extends it?

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
```diff
@@ -534,12 +534,12 @@ public static boolean containsGroup(JsonNode json, String group) {
         repo.setAdminGroup(project.getProjectAdminGroup());
         repo.setUserGroup(project.getProjectUserGroup());
       } else {
-        repo.setAdminGroup(globalKeyuserRoleName);
+        repo.setAdminGroup(defaultAdminGroup);
```
stitakis, line 537: This seems to not be covered with a unit test. Can you please extends it?

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
jafarre-viewnext: If we do that, the default admin group will also get admin rights on the project itself. This should be agreed by Irene.

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
jafarre-viewnext: Well, I thought that passing an independent flag could be error prone or misinterpreted to allow customization. It also feels like showing out the implementation. If, in the future, we changed the criteria to assign the permissions, the flag could become useless and additional parameters may be needed.
That's why I  decided to pass the project information as a whole, even if we are currently only using the key and that flag.
What do you think?

jafarre-viewnext:
> @jafarre-viewnext I think is time to document the implemented rules. Proper documentation for this is missing and causes recurring queries on how the provision app behaves. Can you please extends https://github.com/opendevstack/ods-provisioning-app/blob/master/docs/modules/provisioning-app/pages/index.adoc to capture the new implemented rules?

Done

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
jafarre-viewnext: Done

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
jafarre-viewnext: Done

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
stitakis: If we need to change the criteria the in the future, the code will be changed anyway. In this case having a method signature that explicitly defines what it uses makes the code easier to read. I think, readability wins here.

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
```diff
@@ -477,14 +477,14 @@ public static boolean containsGroup(JsonNode json, String group) {
           repo.setAdminGroup(project.getProjectAdminGroup());
```
stitakis, line 477: This part seems now to not be covered with a unit test. Can you please extend it?

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
```diff
@@ -534,12 +534,12 @@ public static boolean containsGroup(JsonNode json, String group) {
         repo.setAdminGroup(project.getProjectAdminGroup());
```
stitakis, line 534: This seems now to not be covered with a unit test. Can you please extend it?

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
stitakis: > the default admin group will also get admin rights on the project itself. 

I noticed that is happening anyway:  See line method `createComponentRepositoriesForODSProject`:
https://github.com/opendevstack/ods-provisioning-app/blob/544ce527c6101fa238eff5e249b634b9d6765277/src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java#L476-L482

stitakis (changes requested):
Good documentation, thanks! The PR still need some more unit test and there are some comments to clarify

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
jafarre-viewnext: As we have agreed, this idea is discarded due to backward compatibility issues.

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
jafarre-viewnext: Done

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
jafarre-viewnext: Done

src/main/java/org/opendevstack/provision/services/BitbucketAdapter.java
jafarre-viewnext: Ok, done

stitakis (approved):
Well done! Thanks!

PR #703, opened 2021-11-24
API: fix bug on remove quickstarters (deleted quickstarter still returned) - master

### LINKED ISSUE #702: DELETE_COMPONENTS API stores and returns project data with deleted quickstarter
**Describe the bug**

**To Reproduce**
- provision a project and a component
- use the api with delete_componet - and notice the quickstarter (although deleted) is still in the returned list

**Expected behavior**
The project structure returned does NOT contain the deleted quickstarter anymore (only the repo which we DO NOT delete by design)

**Affected version (please complete the following information):**
 - OpenShift: [3/4]
 - OpenDevStack [all versions, 3x++]

**Log Output (ensure to remove any confidential information like tokens, project names, etc.**
delete_componet request:
```
[2021-11-24T06:00:41.707Z] [0;32m    amazon-ebs: {[0m
[2021-11-24T06:00:41.707Z] [0;32m    amazon-ebs:   "projectKey" : "ODSVERIFY",[0m
[2021-11-24T06:00:41.707Z] [0;32m    amazon-ebs:   "quickstarters" : [[0m
[2021-11-24T06:00:41.707Z] [0;32m    amazon-ebs:     {[0m
[2021-11-24T06:00:41.707Z] [0;32m    amazon-ebs:       "component_type" : "docker-plain",[0m
[2021-11-24T06:00:41.707Z] [0;32m    amazon-ebs:       "component_id" : "plain"[0m
[2021-11-24T06:00:41.707Z] [0;32m    amazon-ebs:     }[0m
[2021-11-24T06:00:41.707Z] [0;32m    amazon-ebs:   ][0m
[2021-11-24T06:00:41.707Z] [0;32m    amazon-ebs: }[0m
``` 

result from the api call
```
"quickstarters":
	[{"component_type":"docker-plain","component_id":"plain",
		"git_url_ssh":"ssh://git@bitbucket.odsbox.lan:7999/odsverify/odsverify-plain.git",
		"git_url_http":"http://bitbucket.odsbox.lan:7990/scm/odsverify/odsverify-plain.git",
		"GROUP_ID":"org.opendevstack.odsverify",
		"PROJECT_ID":"odsverify",
		"PACKAGE_NAME":"org.opendevstack.odsverify.plain",
		"ODS_NAMESPACE":"ods",
		"ODS_BITBUCKET_PROJECT":"opendevstack","ODS_IMAGE_TAG":"latest","ODS_GIT_REF":"4.x"}]	
```

clemensutschig:
closes #702 on master

### DIFF
--- a/src/main/java/org/opendevstack/provision/model/OpenProjectData.java
+++ b/src/main/java/org/opendevstack/provision/model/OpenProjectData.java
@@ -125,14 +125,14 @@ public class OpenProjectData {
       if (quickStarterToRemove.get(COMPONENT_ID_KEY) == null) {
         throw new NullPointerException("Cannot delete quickstarter with id null!");
       }
-    }
-    for (Map<String, String> quickStarterToRemove : quickstartersToRemove) {
+      // loop over the currently provisioned quickstarters, and find the
+      // one with similar id - to then remove it.
       List<Map<String, String>> currentQuickstarters = getQuickstarters();
       for (int i = 0; i < currentQuickstarters.size(); i++) {
         Map<String, String> currentQuickstarter = currentQuickstarters.get(i);
         if (currentQuickstarter
             .get(COMPONENT_ID_KEY)
-            .equals(quickStarterToRemove.get(COMPONENT_ID_KEY))) {
+            .equalsIgnoreCase(quickStarterToRemove.get(COMPONENT_ID_KEY))) {
           quickstarters.remove(i);
         }
       }
--- a/src/main/java/org/opendevstack/provision/model/OpenProjectData.java
+++ b/src/main/java/org/opendevstack/provision/model/OpenProjectData.java
@@ -121,8 +121,21 @@ public class OpenProjectData {
     if (quickstartersToRemove == null) {
       return this.quickstarters;
     }
-    for (Map<String, String> quickstarter : quickstartersToRemove) {
-      quickstarters.remove(quickstarter);
+    for (Map<String, String> quickStarterToRemove : quickstartersToRemove) {
+      if (quickStarterToRemove.get(COMPONENT_ID_KEY) == null) {
+        throw new NullPointerException("Cannot delete quickstarter with id null!");
+      }
+    }
+    for (Map<String, String> quickStarterToRemove : quickstartersToRemove) {
+      List<Map<String, String>> currentQuickstarters = getQuickstarters();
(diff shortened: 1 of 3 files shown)


clemensutschig:
also in spirit of good doc - 

the fix is in the `removeQuickstarterFromProject` method - which used a `hash remove` instead of an id-based remove. This is obviously wrong, because the quickstarter removal API only needs an `id`, and that's it. The reason why this did not show up in the tests is because of a bug there as well. For removal it used the full quickstarter structure retrieved from getProjects instead of a request with just the `id`

stitakis:
@clemensutschig I'll to review today, ideally in the morning

src/main/java/org/opendevstack/provision/model/OpenProjectData.java
```diff
@@ -121,8 +121,21 @@ public int hashCode() {
...
+      for (int i = 0; i < currentQuickstarters.size(); i++) {
+        Map<String, String> currentQuickstarter = currentQuickstarters.get(i);
+        if (currentQuickstarter
+            .get(COMPONENT_ID_KEY)
+            .equalsIgnoreCase(quickStarterToRemove.get(COMPONENT_ID_KEY))) {
+          quickstarters.remove(i);
+        }
+      }
     }
     return quickstarters;
```
stitakis, line 140: After a quick review I found some design issues with this class, i.e. it break encapsulation of internals by exposing the quickstarters map to the method caller. This could modify the quickstarters map outside the boundaries of the OpenProjectData class. Also removal of quickstarters is not thread safe. However, I'm concentrating here on fix for the reported bug...
I would suggest that we modify the method "removeQuickstartersFromProject" to not return the quickstarters map. Simply because the method name does make clear what is returning. It has currently only one caller which it doesn't need the quickstarters map. Also from the caller point of view is not clear if a list of the removed quickstarters or the remaining quickstarters are returned. A developer will need to review the implementation to understand what this method is returning. Alternatively we could rename the method to reflect its intend (e.g. "removeQuickstartersFromProjectAndReturnRemainingOnes") but I'm would prefer to remove the return clauses. Sounds good?

clemensutschig: @stitakis - I changed it now, to return the removed quickstarters. Happier?

stitakis (approved):

PR #705, opened 2021-11-26
fix wrong execution job on DELETE API - for master

clemensutschig:
closes jenkins shared lib #790 (https://github.com/opendevstack/ods-jenkins-shared-library/issues/790) for master

### DIFF
--- a/src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
+++ b/src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
@@ -538,7 +538,7 @@ public class JenkinsPipelineAdapter extends BaseServiceAdapter implements IJobEx
 
       CLEANUP_LEFTOVER_COMPONENTS objectType = CLEANUP_LEFTOVER_COMPONENTS.PLTF_PROJECT;
       // Note: the secret passed here is the corresponding to the ODS CD webhook proxy
-      return runDeleteAdminJob(
+      return runDeleteAdminJobAndSetAsLastExecutionJobToProject(
           deleteProjectAdminJob,
           project.getProjectKey(),
           openshiftJenkinsTriggerSecret,
@@ -559,7 +559,7 @@ public class JenkinsPipelineAdapter extends BaseServiceAdapter implements IJobEx
             .map(q1 -> q1.get(OpenProjectData.COMPONENT_ID_KEY))
             .map(
                 component ->
-                    runDeleteAdminJob(
+                    runDeleteAdminJobAndSetAsLastExecutionJobToProject(
                         jenkinsPipelineProperties.getDeleteComponentsQuickstarter(),
                         project.getProjectKey(),
                         project.getWebhookProxySecret(), // Note: the secret passed here is the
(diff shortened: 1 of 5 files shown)


src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
```diff
@@ -592,8 +595,18 @@ private static void validateComponentName(
       logger.debug("Calling job {} for project {}", job.getId(), projectKey);
       ExecutionsData data = prepareAndExecuteJob(job, options, webhookProxySecret);
       logger.info("Result of cleanup: {}", data.toString());
+      ExecutionJob result = new ExecutionJob();
+      result.setName(data.getJobName());
+      result.setUrl(data.getPermalink());
+      List<ExecutionJob> currentJobs = project.getLastExecutionJobs();
+      if (currentJobs == null) {
+        currentJobs = new ArrayList<ExecutionJob>();
+        project.setLastExecutionJobs(currentJobs);
```
stitakis, line 604: This line changes the object `project`. It is not clear from the caller method point of view the effect on the object roject. Therefore it is a side effect which generally should be avoid.

I'd tend to put the logic that modify object `project` in the `ProjectAPIController`. To do this, some few more changes would need to be done, like modifying the interface method  `IServiceAdapter.cleanUp(...)` to give back the execution job details.
 
If this change is urgent then I'd suggest at least to rename the method to express the side-effect in the object `project`. The name would be then e.g. : `runDeleteAdminJobAndSetAsLastExecutionJobToProject`

stitakis: Let us reduce the number of code lines to try  to make it easier to read, e.g. to:

```
      ExecutionJob result = new ExecutionJob();
      result.setName(data.getJobName());
      result.setUrl(data.getPermalink());

      if (project.getLastExecutionJobs() == null) {
        project.setLastExecutionJobs(new ArrayList<ExecutionJob>());
      }
      project.getLastExecutionJobs().add(result);
```
Looks better?

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
```diff
@@ -592,8 +595,18 @@ private static void validateComponentName(
...
+      }
+      currentJobs.add(result);
       return Collections.emptyMap();
     } catch (RuntimeException | IOException e) {
+      e.printStackTrace();
```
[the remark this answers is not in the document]
stitakis: done

src/main/java/org/opendevstack/provision/services/JenkinsPipelineAdapter.java
stitakis: I've renamed the method to the suggested name `runDeleteAdminJobAndSetAsLastExecutionJobToProject` and also reorder the code to make it more readable

stitakis (approved):

clemensutschig:
4.x yes - 3.x no - imho

stitakis:
ok, I'll backport this tomorrow to `4.x`

PR #707, opened 2021-12-02
feature: enable a configurable disclaimer in the UI with a new property

### LINKED ISSUE #706: Add configurable disclaimer text in the provisioning app
As we need to show a disclaimer to our users, we would need to add a configurable disclaimer text to be shown in all the pages.

Solution proposed:
Add a new property called "provision.ui.disclaimer" that can be empty or a text with links inside.
In the header template, we check if it's informed and in that case we display a new section with the message inside.

jordivx:
Resolves #706 
Added a property to define the disclaimer text. And a text in the header component to be displayed in all the screens in case it is defined.


stitakis (approved):

metmajer:
Well done @jordivx 👍

Issue #720, opened 2022-03-01
Openshift service adapter is timing out with openshift 4


stitakis:
The openshift service adapter is timing out on API calls with red hat dedicated oc4.

The current used openshift library uses openshift 3.x API, which is probably is not enabled anymore in dedicated oc4.

While this needs to be confirmed, the solution to this problem would be an upgrade of the library to support oc4 API.
However, we should decide if we want to maintain support for oc3. If this is the case, we would need a way to provide to openshift adapters and import the corresponding libraries.

As first measure we should change the behavior to disable this service by default (#721)

clemensutschig:
did you folks try with the OCP4 libs, and that works? - just asking :) should be easily reproducable locally, or?

oalyman:
I think you need to use this as an alternative https://github.com/fabric8io/kubernetes-client/

stitakis:
I haven't used the ocp4 libs yet... I'm currently focusing on other topics... if you time to fix, I'm happy to give a hand

Issue #721, opened 2022-03-01
Disable openshift service adapter by default


stitakis:
To avoid problem in the corporate environment while this issue is open #720 we will disable the openshift service adapter by default

PR #727, opened 2022-03-04
update openshift client to use fabric8 openshift client (openshift 4 …compatible)

oalyman:
Fixes: #720 and rolles back #721


docs/modules/provisioning-app/pages/configuration.adoc
```diff
@@ -413,10 +413,10 @@ To disable the confluence adapter you can add this property to the application p
 adapters.confluence.enabled=false
 ```
 
-The Openshift Service can be used to verify that a project key does not exist in the cluster before provisioning a project.
-If you want to enable it, you can add this property to the application properties:
+The Openshift Service currently is used to verify that a project key does not exist in the cluster before provisioning a project.
+If you need to disable it, you can add this property to the application properties:
 ```
-services.openshift.enabled=true
+services.openshift.enabled=false
```
stitakis, line 419: I would keep this feature disabled until we have the chance to test it in openshift, or did you have the chance to deploy this version of prov app in openshift?

stitakis (commented):
top! Thanks a lot for fixing this! I'd approve it but I'm not sure whether we should enable the service at this moment. See my inline comment.

docs/modules/provisioning-app/pages/configuration.adoc
stitakis: excellent! Let us merge this change and the ODS build run the E2E tests. Then we can verify that nothing breaks before porting this change to branch `4.x`. Sounds good?

stitakis (approved):
Let us merge this change and the ODS build run the E2E tests. Then we can verify that nothing breaks before porting this change to branch 4.x. Sounds good?

PR #735, opened 2022-05-05
Shortcuts creation removed (EDPC-1288)

fbeba-bi:


src/main/java/org/opendevstack/provision/controller/ProjectApiController.java
```diff
@@ -251,9 +251,6 @@ public ResponseEntity<Object> addProject(
       // create the delivery chain, including scm repos, and platform project
       newProject = createDeliveryChain(newProject);
```
victorpablosceruelo, line 253: Maybe you should remove the method in addition to the removal of this call to it...
Is it possible to remove the method or it is used somewhere else?

src/main/java/org/opendevstack/provision/controller/ProjectApiController.java
```diff
@@ -251,9 +251,6 @@ public ResponseEntity<Object> addProject(
...
 
-      // add shortcuts into the space
-      jiraAdapter.addShortcutsToProject(newProject);
```
victorpablosceruelo, line 255: Maybe you should remove the method in addition to the removal of this call to it...
Is it possible to remove the method or it is used somewhere else?

hrcornejo (approved):

